PR #18, opened 2024-04-05
Remove unused traits

cwpearson:
Remove some dead code

### DIFF
--- a/src/KokkosComm_traits.hpp
+++ b/src/KokkosComm_traits.hpp
@@ -49,18 +49,6 @@ struct Traits<View> {
   static size_t extent(const View &v, const int i) { return v.extent(i); }
   static size_t stride(const View &v, const int i) { return v.stride(i); }
 
-  template <typename ExecSpace>
-  static void pack(const ExecSpace &space,
-                   const non_const_packed_view_type &dst, const View &src) {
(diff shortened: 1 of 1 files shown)


src/KokkosComm_traits.hpp
```diff
@@ -51,18 +51,6 @@ struct Traits<View> {
...
-                   const non_const_packed_view_type &dst, const View &src) {
-    Kokkos::deep_copy(space, dst, src);
-  }
-
-  template <typename ExecSpace>
-  static void unpack(const ExecSpace &space, const View &dst,
-                     const non_const_packed_view_type &src) {
-    Kokkos::deep_copy(space, dst, src);
-  }
-
```
masterleinad, line 63: Why exactly are you deleting all of this?

cwpearson: All the packing got moved over here https://github.com/cwpearson/kokkos-comm/blob/develop/src/impl/KokkosComm_packer.hpp. There's a deep_copy implementation in there and also one that constructs an MPI Datatype and passes it off to MPI.

I must have forgotten to remove these when I did that. They're currently unused.

cwpearson:
Since `mdspan` is gone, this now just removes some dead code.

cedricchevalier19 (approved):

PR #26, opened 2024-04-09
Remove mdspan support

### LINKED ISSUE #25: mdspan support when Kokkos already has one
If Kokkos has been installed with mdspan support, i.e. `Kokkos_ENABLE_IMPL_MDSPAN=ON`, none of the configuration works.

Adding a new `mdspan` results in duplicated symbols given `mdspan` headers are protected with `#pragma once` and the header can be found twice.

Disabling the `FetchContent` for `mdspan` does not work either.

masterleinad:
Fixes #25. We discussed that we don't want `mdspan` support for now.


cedricchevalier19 (commented):
I am in favour of this PR as it will remove the burden to maintain a long term and experimental feature.

For research experiments, one can restore the previous code (that is a good starting point for mdspan) from git.

cedricchevalier19 (approved):
I let @cwpearson do the merge if he agrees.

cwpearson:
Thanks for this!

Issue #10, opened 2024-03-26
Implement Teuchos MPI operations


cwpearson:
The motivation here is to prove we have a reasonable implementation by being able to run Trilinos on top of KokkosComm

| MPI_... | 1D | non-contiguous |
|-|-|-|
| allreduce | |
| broadcast | |
| gather | |
| scan | |

- [ ] abort
- [x] allgather
- [x] barrier
- [x] irecv
  - [ ] non-contiguous [\#32](https://github.com/kokkos/kokkos-comm/pull/32)
- [x] rsend
- [x] recv
- [x] send
- [x] ssend

cedricchevalier19:
Very good starting point.

I have a comment concerning the API. Here we have 3 "blocking" `send` and only one "immediate". 
It might make sense to have only one of each and a template parameter to choose between the implementation.

```C++
send(snd_buf, comm); // Alias for send<Default>(snd_buff, comm);
send<Ready>(snd_buf, comm);
send<Synchronous>(snd_buf, comm);
send<Buffered>(snd_buf, comm);
```

This way we can also implement the `isend` variants. I think we should try to push users to `I` variants as they are a lot easier to use to make a semantically sound code, and also be ready for more asynchronous computations.

cwpearson:
That's definitely an option, but since the `Immediate` version returns `KokkosComm::Req` and the rest do not, you actually have to do something like:

```c++
enum class Mode {
    Blocking,
    Buffered,
    Immediate,
    Ready,
    Synchronous
};

// most modes just return an error code
template <Mode mode>
struct SendReturn {
    using type = int;
}

// Immediate return value includes an MPI_Request
struct IsendResult {
    int err;
    MPI_Request req;
};
template<> struct SendReturn<Mode::Immediate> {
    using type = IsendResult;
}

template <Mode mode>
typename SendReturn<mode>::type send(...);

template<> typename SendReturn<Blocking>::type send<Blocking>(...) {...}
...
```

I think (though I might be missing something), this basically just replaces symbol lookup with template resolution in the compiler, and I'm not sure it really buys us anything

cwpearson:
I see that I misunderstood a bit - you were just suggesting doing this for the synchronous versions. I think my point still stands (though we can avoid the funky return type template).

cedricchevalier19:
I think `send` and `isend` should be separate functions as they are semantically very different.
But between variants they are quite interchangeable.

[...]

PR #30, opened 2024-04-11
Add communication modes to specialize send routines

dssgabriel:
This PR adds an enum to allow users to specify the communication mode of send routines (both blocking and non-blocking) as proposed by @cedricchevalier19 in #10. The `CommMode` template parameter is only used for the user-facing API: there are dedicated lean wrappers for each underlying MPI call in the `Impl` namespace (as suggested by @cwpearson, also in #10). Relevant unit tests have been added too.

The changes fulfill the requirements for supporting `rsend` and `ssend` from #10.

We do not support "Buffered" mode (at least for now), as it requires the user to manually attach a buffer using `MPI_Buffer_attach` (see documentation [here](https://docs.open-mpi.org/en/v5.0.x/man-openmpi/man3/MPI_Bsend.3.html#notes)).


dssgabriel:
We may want to draft this for now as I haven't updated the corresponding documentation yet.

unit_tests/test_irsendrecv.cpp
```diff
@@ -0,0 +1,94 @@
...
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <gtest/gtest.h>
+
+#include "KokkosComm.hpp"
+
```
cedricchevalier19: Perhaps we should put all the tests in an anonymous namespace

unit_tests/test_rsendrecv.cpp
```diff
@@ -0,0 +1,86 @@
...
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <gtest/gtest.h>
+
+#include "KokkosComm.hpp"
+
```
cedricchevalier19: Perhaps we should put all the tests in an anonymous namespace.

src/KokkosComm_mode.hpp
```diff
@@ -0,0 +1,29 @@
...
+//
+//@HEADER
+
+#pragma once
+
+#include <cstdint>
+
+namespace KokkosComm {
+
+enum class Mode : uint8_t {
```
cedricchevalier19: Why this `uint8_t` specification?

And `Mode` is kind of vague.

src/KokkosComm_mode.hpp
```diff
@@ -0,0 +1,29 @@
...
+  Default,
```
cedricchevalier19: Describe these values.

src/KokkosComm.hpp
```diff
@@ -22,21 +22,36 @@
 #include "KokkosComm_recv.hpp"
 #include "KokkosComm_send.hpp"
 #include "KokkosComm_concepts.hpp"
+#include "KokkosComm_mode.hpp"
```
cedricchevalier19: Perhaps we can avoid this mode.hpp as it is mostly empty.

src/KokkosComm.hpp
```diff
@@ -22,21 +22,36 @@
...
 
 #include <Kokkos_Core.hpp>
 
 namespace KokkosComm {
 
-template <KokkosExecutionSpace ExecSpace, KokkosView SendView>
+template <Mode CommMode = Mode::Default, KokkosExecutionSpace ExecSpace,
```
cedricchevalier19:
```suggestion
template <KokkosExecutionSpace ExecSpace,KokkosView SendView,
          Mode CommMode = Mode::Default>
```

src/KokkosComm.hpp
```diff
@@ -22,21 +22,36 @@
...
+    return Impl::isend(space, sv, dest, tag, comm);
+  } else if constexpr (CommMode == Mode::Ready) {
+    return Impl::irsend(space, sv, dest, tag, comm);
+  } else if constexpr (CommMode == Mode::Synchronous) {
+    return Impl::issend(space, sv, dest, tag, comm);
+  }
 }
 
-template <KokkosExecutionSpace ExecSpace, KokkosView SendView>
+template <Mode CommMode = Mode::Default, KokkosExecutionSpace ExecSpace,
```
cedricchevalier19:
```suggestion
template <KokkosExecutionSpace ExecSpace,KokkosView SendView,
          Mode CommMode = Mode::Default>
```

cedricchevalier19 (commented):
Just a quick look, but I think it makes more sense to group all the sendrecv tests in the same file.

You can define a function that take the comm "mode" as a template parameters and have tests that simplify set this parameter.

src/KokkosComm_mode.hpp
dssgabriel: I propose `CommMode` for the name then. It's more explicit and remains short enough.

As for the `uint8_t` specification, it's just to ensure the enum is defined using a narrow type, as ISO C++ defaults to `int` (see section 9.7.1.5 of the [latest draft of the standard](https://open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4950.pdf)), which wastes space for such a small set of variants.

src/KokkosComm.hpp
dssgabriel: Best to move the `Mode` enum definition to the main `KokkosComm` header then?

src/KokkosComm_mode.hpp
cedricchevalier19: I do not think we care that much as it is for compile time selection.

dssgabriel: I'm fine with removing it then. :+1:

docs/api/core.rst
```diff
@@ -41,11 +49,12 @@ Point-to-point
     :tparam SendView: A Kokkos::View to send
     :tparam ExecSpace: A Kokkos execution space to operate in
     :returns: A KokkosComm::Req representing the asynchronous communication and any lifetime-extended views.
+    :tparam SendMode: A communication mode to use. If unspecified, defaults to a standard ``MPI_Isend``.
```
cedricchevalier19: Is it possible to add a link to the enum of valid values?

src/impl/KokkosComm_send.hpp
```diff
@@ -42,4 +43,43 @@ void send(const ExecSpace &space, const SendView &sv, int dest, int tag,
...
+    space.fence();
+    MPI_Ssend(args.view.data(), args.count, args.datatype, dest, tag, comm);
+  } else {
+    using SendScalar = typename SendView::value_type;
+    MPI_Ssend(sv.data(), sv.span(), mpi_type_v<SendScalar>, dest, tag, comm);
+  }
+
+  Kokkos::Tools::popRegion();
+}
+
```
cedricchevalier19: Not convinced by the code duplication.
It can be nice to have separate functions, but as the code is strictly the same except for the `MPI_?send` call, I think factorizing the code and doing the dispatch here make more sense.

dssgabriel: Wouldn't this require including a public header into a private one?
As `KokkosComm::CommMode` is defined in `KokkosComm.hpp`, we would have a circular dependency. Any ideas on how I can manage that?

docs/api/core.rst
dssgabriel: Do you mean to add a hyperlink target to the `KokkosComm::CommMode` section in the documentation?

dssgabriel:
The PR now does the communication mode dispatch in the `Impl` namespace to avoid code duplication.

It now also lets users override the default communication mode to alias `CommMode::Synchronous` instead of `CommMode::Standard` (mainly for debugging purposes) using the `KOKKOSCOMM_FORCE_SYNCHRONOUS_SEND_MODE` environment variable.

docs/api/core.rst
```diff
@@ -28,10 +36,10 @@ Core
 Point-to-point
 --------------
 
-.. cpp:function:: template <KokkosExecutionSpace ExecSpace, KokkosView SendView> \
+.. cpp:function:: template <KokkosComm::CommMode SendMode = CommMode::Standard, KokkosExecutionSpace ExecSpace, KokkosView SendView> \
```
cedricchevalier19: Not sure about putting the default value in the prototype in the doc.

docs/api/core.rst
```diff
@@ -41,11 +49,13 @@ Point-to-point
     :tparam SendView: A Kokkos::View to send
     :tparam ExecSpace: A Kokkos execution space to operate in
     :returns: A KokkosComm::Req representing the asynchronous communication and any lifetime-extended views.
+    :tparam SendMode: A communication mode to use. If unspecified, defaults to a standard ``MPI_Isend``.
```
cedricchevalier19: Not always the "standard" if we add a "debug" feature.

docs/api/core.rst
```diff
@@ -90,6 +102,29 @@ Collective
...
+
+      Ready mode: Send operations may be started only if the matching receive is already started.
+
+    .. cpp:enumerator:: KokkosComm::CommMode::Synchronous
+
+      Synchronous mode: Send operations complete successfully only if a matching receive is started, and the receive operation has started to receive themessage sent.
+
+    .. cpp:enumerator:: KokkosComm::CommMode::Default
+
+    Default mode is an alias for ``Standard`` mode, but lets users override the behavior of send operations at compile-time using the ``KOKKOSCOMM_FORCE_SYNCHRONOUS_SEND_MODE`` environment variable. This forces ``Synchronous`` mode for all "default-mode" send operations, which can be useful for debugging purposes, e.g., for asserting that the communication scheme is correct.
```
cedricchevalier19: Is it a C++ define or a CMake option?

src/impl/KokkosComm_isend.hpp
```diff
@@ -47,13 +49,45 @@ KokkosComm::Req isend(const ExecSpace &space, const SendView &sv, int dest,
...
+                 tag, comm, &req.mpi_req());
+    } else {
+#ifdef KOKKOSCOMM_FORCE_SYNCHRONOUS_SEND_MODE
+      MPI_Issend(KCT::data_handle(args.view), args.count, args.datatype, dest,
+                 tag, comm, &req.mpi_req());
+#else
+      MPI_Isend(KCT::data_handle(args.view), args.count, args.datatype, dest,
+                tag, comm, &req.mpi_req());
+#endif
+    }
```
cedricchevalier19: Not sure about the `else` as it is already `if constexpr` and the enum cases are clearly disjoint.

docs/api/core.rst
dssgabriel: What do you mean by "C++ define" vs CMake option? Is the former a simple pre-processor definition and the latter an explicit option in the CMake config?

cedricchevalier19: I do not think you use an environment variable named `KOKKOSCOMM_FORCE_SYNCHRONOUS_SEND_MODE` and the `#ifdef` is with `KokkosComm_FORCE_SYNCHRONOUS_MODE` so this part of the documentation is not very clear.

I would rather have a `KOKKOSCOMM_FORCE_SYNCHRONOUS_SEND_MODE` define and a `KokkosComm_FORCE_SYNCHRONOUS_MODE` if it is at the CMake level.

dssgabriel: So the idea would be to have a CMake option `KokkosComm_FORCE_SYNCHRONOUS_MODE` (which we may want to enable when building with the Debug profile) and make it pre-processor define `KOKKOSCOMM_FORCE_SYNCHRONOUS_SENDS` in the C++ code (via adding `-DKOKKOSCOMM_FORCE_SYNCHRONOUS_SENDS` to the flags)?

I am unsure about the wording to use in the documentation, any suggestions?

cedricchevalier19: Leave out, the `CMake` option for now. We have to figure out the best way to handle this flag.

dssgabriel: The pre-processor definition is now `KOKKOSCOMM_FORCE_SYNCHRONOUS_MODE`; we can probably delay adding a corresponding CMake option to another PR.

Are you happy with the current wording in the docs?

docs/api/core.rst
```diff
@@ -2,23 +2,31 @@ Core
...
       - ``Kokkos::View``
-    * - MPI_Send
-      - send
+    * - ``MPI_Send``
+      - ``send`` or ``send<CommMode::Standard>``
       - ✓
+    * - ``MPI_Rsend``
+      - ``send<CommMode::Ready>``
       - ✓
     * - MPI_Recv
```
masterleinad: This should be consistent one way or the other.
```suggestion
    * - ``MPI_Recv``
```

docs/api/core.rst
```diff
@@ -90,6 +100,29 @@ Collective
...
+
+      Standard mode: the MPI implementation decides whether outgoing messages will be buffered. Send operations can be started whether or not a matching receive has been started. They may complete before a matching receive is started. Standard mode is non-local: successful completion of the send operation may depend on the occurrence of a matching receive.
+
+    .. cpp:enumerator:: KokkosComm::CommMode::Ready
+
+      Ready mode: Send operations may be started only if the matching receive is already started.
+
+    .. cpp:enumerator:: KokkosComm::CommMode::Synchronous
+
+      Synchronous mode: Send operations complete successfully only if a matching receive is started, and the receive operation has started to receive themessage sent.
```
masterleinad:
```suggestion
      Synchronous mode: Send operations complete successfully only if a matching receive is started, and the receive operation has started to receive the message sent.
```

src/KokkosComm_comm_mode.hpp
```diff
@@ -0,0 +1,43 @@
...
+#pragma once
+
+namespace KokkosComm {
+
+// Scoped enumeration to specify the communication mode of a sending operation.
+// See section 3.4 of the MPI standard for a complete specification.
+enum class CommMode {
+  // Default mode: lets the user override the send operations behavior at
+  // compile-time. E.g., this can be set to mode "Synchronous" for debug
+  // builds.
```
masterleinad:
```suggestion
  // Default mode: lets the user override the send operations behavior at
  // compile-time. E.g., this can be set to mode "Synchronous" for debug
  // builds by defining KOKKOSCOMM_FORCE_SYNCHRONOUS_MODE.
```

src/impl/KokkosComm_isend.hpp
```diff
@@ -47,13 +49,45 @@ KokkosComm::Req isend(const ExecSpace &space, const SendView &sv, int dest,
...
+    } else if constexpr (SendMode == CommMode::Default) {
+#ifdef KOKKOSCOMM_FORCE_SYNCHRONOUS_MODE
+      MPI_Issend(KCT::data_handle(args.view), args.count, args.datatype, dest,
+                 tag, comm, &req.mpi_req());
+#else
+      MPI_Isend(KCT::data_handle(args.view), args.count, args.datatype, dest,
+                tag, comm, &req.mpi_req());
+#endif
+    }
     req.keep_until_wait(args.view);
```
masterleinad: I'd rather define a lambda so you don't have to repeat the logic for tyes that need packing or not.

src/impl/KokkosComm_send.hpp
```diff
@@ -34,12 +37,37 @@ void send(const ExecSpace &space, const SendView &sv, int dest, int tag,
...
+      MPI_Rsend(args.view.data(), args.count, args.datatype, dest, tag, comm);
+    } else if constexpr (SendMode == CommMode::Synchronous) {
+      MPI_Ssend(args.view.data(), args.count, args.datatype, dest, tag, comm);
+    } else if constexpr (SendMode == CommMode::Default) {
+#ifdef KOKKOSCOMM_FORCE_SYNCHRONOUS_MODE
+      MPI_Ssend(args.view.data(), args.count, args.datatype, dest, tag, comm);
+#else
+      MPI_Send(args.view.data(), args.count, args.datatype, dest, tag, comm);
+#endif
+    }
```
masterleinad: I'd rather define a lambda so you don't have to repeat the logic for tyes that need packing or not.

dssgabriel:
@masterleinad Can you tell me if the lambda definitions look ok? Using `auto` everywhere seems to work, but I am not very proficient in C++ and not sure this is the best way to do it... :sweat_smile: 
Also, do I need to mark the lambdas as `constexpr` to ensure the path is correctly selected at compile-time?

src/impl/KokkosComm_isend.hpp
```diff
@@ -40,20 +42,35 @@ KokkosComm::Req isend(const ExecSpace &space, const SendView &sv, int dest,
   using KCT  = KokkosComm::Traits<SendView>;
   using KCPT = KokkosComm::PackTraits<SendView>;
 
+  auto isend_fn = [&](auto &&view, auto count, auto datatype, auto &&request) {
```
cedricchevalier19: I do not think you need to capture a lot of external things.
Perhaps you can explicitly list what you capture or still pass `dest`, `tag`, and `comm` as arguments?

masterleinad: I would also prefer being explicit about the types that are fixed. Also, I don't think we need universal references here but can just pass Views by value or reference explicitly.

dssgabriel: I am changing the lambda prototypes to match the ones of MPI functions, with explicit typing.

masterleinad (dismissed):
Looks OK to me.

unit_tests/test_isendrecv.cpp
```diff
@@ -85,10 +86,40 @@ TYPED_TEST(IsendRecv, 1D_noncontig) {
...
   }
 }
+
+TYPED_TEST(IsendRecv, 1D_contig_standard) {
+  isend_comm_mode_1d_contig<KokkosComm::CommMode::Standard,
+                            typename TestFixture::Scalar>();
+}
+
+TYPED_TEST(IsendRecv, 1D_contig_ready) {
+  isend_comm_mode_1d_contig<KokkosComm::CommMode::Ready,
```
devreal, line 109: The ready-mode test is incorrect. because it does not ensure that the receiver has actually posted a matching receive. Most implementations will handle it gracefully but it may fail spuriously if you're unlucky (plus, the test should show correct usage of `KokkosComm::CommMode::Ready`).

dssgabriel: I believe this would require using `irecv` instead of `recv` then?   
Or is there another way to ensure a Ready-mode send is semantically correct other than the following?
```cpp
if (rank == 0) {
  MPI_Barrier(MPI_COMM_WORLD);
  MPI_Rsend(/* args... */);
} else {
  MPI_Irecv(/* args... */, &req);
  MPI_Barrier(MPI_COMM_WORLD);
  MPI_Wait(&req, MPI_STATUS_IGNORE);
}
```
This would require merging #32 to replace the blocking `recv` with an immediate one and maybe #40 for `KokkosComm::barrier`.

dssgabriel:
I rebased the latest commits from `develop` onto this branch.

After discussing with @cedricchevalier19, we believe it's simpler to skip tests for ready-mode send operations for now. We should merge this PR and open an issue to fix ready-send tests while we wait for #32 to be merged.

unit_tests/test_isendrecv.cpp
devreal: Yes, you need a barrier here. I'm fine with skipping this test.

cedricchevalier19 (approved):

cwpearson:
Works for me, sorry for the delay on https://github.com/kokkos/kokkos-comm/pull/32. Open such an issue and then we will merge once conflicts are resolved!

dssgabriel:
Issue #43 for the skipped tests is opened. I think we can merge this now :+1:

devreal (approved):

PR #39, opened 2024-04-22
mpi: contiguous MPI_Allgather

cwpearson:
Leave non-contiguous cases for when our approach is more resolved.

### DIFF
--- a/unit_tests/test_allgather.cpp
+++ b/unit_tests/test_allgather.cpp
@@ -1,0 +1,82 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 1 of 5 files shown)


unit_tests/test_allgather.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  KokkosComm::allgather(Kokkos::DefaultExecutionSpace(), sv, rv,
+                        MPI_COMM_WORLD);
+
+  int errs;
+  Kokkos::parallel_reduce(
+      rv.extent(0),
+      KOKKOS_LAMBDA(const int &i, int &lsum) {
+        const int src = i / nContrib;
+        const int j   = i % nContrib;
```
cedricchevalier19, line 77:
```suggestion
        const int j   = i  - src * nContrib;
```

cedricchevalier19 (dismissed):

unit_tests/test_allgather.cpp
cwpearson: it's not performance-sensitive and I think the first one is a pretty clear idiom for what's going on

cedricchevalier19: Ok, I think I may have been traumatised by https://cplusplus.github.io/CWG/issues/614.html when having to deal with signed numbers.

cedricchevalier19 (dismissed):

cedricchevalier19 (approved):

PR #48, opened 2024-05-08
More MPI Datatypes

dutkalex:
Title says it all :)

### DIFF
--- a/src/impl/KokkosComm_types.hpp
+++ b/src/impl/KokkosComm_types.hpp
@@ -28,6 +28,19 @@ MPI_Datatype mpi_type() {
 }
 
 template <>
+inline MPI_Datatype mpi_type<char>() {
+  return MPI_CHAR;
+}
+template <>
+inline MPI_Datatype mpi_type<unsigned char>() {
+  return MPI_UNSIGNED_CHAR;
+}
+template <>
+inline MPI_Datatype mpi_type<std::byte>() {
+  return MPI_BYTE;
+}
+
+template <>
 inline MPI_Datatype mpi_type<short>() {
   return MPI_SHORT;
 }
--- a/src/impl/KokkosComm_types.hpp
+++ b/src/impl/KokkosComm_types.hpp
@@ -21,131 +21,61 @@
 #include <Kokkos_Core.hpp>
 
 namespace KokkosComm::Impl {
-template <typename Scalar>
-MPI_Datatype mpi_type() {
-  static_assert(std::is_void_v<Scalar>, "mpi_type not implemented");
-  return MPI_CHAR;  // unreachable
-}
-
-template <>
-inline MPI_Datatype mpi_type<char>() {
-  return MPI_CHAR;
-}
-template <>
-inline MPI_Datatype mpi_type<unsigned char>() {
-  return MPI_UNSIGNED_CHAR;
-}
-template <>
-inline MPI_Datatype mpi_type<std::byte>() {
-  return MPI_BYTE;
-}
-
-template <>
-inline MPI_Datatype mpi_type<short>() {
-  return MPI_SHORT;
-}
-template <>
-inline MPI_Datatype mpi_type<int>() {
-  return MPI_INT;
-}
-template <>
-inline MPI_Datatype mpi_type<long int>() {
-  return MPI_LONG;
-}
-template <>
-inline MPI_Datatype mpi_type<long long>() {
-  return MPI_LONG_LONG;
-}
-
-template <>
-inline MPI_Datatype mpi_type<unsigned short>() {
-  return MPI_UNSIGNED_SHORT;
-}
-template <>
-inline MPI_Datatype mpi_type<unsigned int>() {
-  return MPI_UNSIGNED;
-}
-template <>
-inline MPI_Datatype mpi_type<unsigned long>() {
-  return MPI_UNSIGNED_LONG;
-}
-template <>
-inline MPI_Datatype mpi_type<unsigned long long>() {
-  return MPI_UNSIGNED_LONG_LONG;
-}
-
-template <>
-inline MPI_Datatype mpi_type<float>() {
-  return MPI_FLOAT;
-}
-template <>
-inline MPI_Datatype mpi_type<double>() {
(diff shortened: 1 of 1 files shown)


cwpearson:
@dutkalex we require clang-format-8 at present - may be easiest to copy-paste the diff from the failed test and apply it to your branch 😄

src/impl/KokkosComm_types.hpp
```diff
@@ -43,18 +39,33 @@ template <>
...
+  return MPI_UNSIGNED;
 }
 template <>
-inline MPI_Datatype mpi_type<double>() {
-  return MPI_DOUBLE;
+inline MPI_Datatype mpi_type<unsigned long>() {
+  return MPI_UNSIGNED_LONG;
 }
+template <>
+inline MPI_Datatype mpi_type<unsigned long>() {
```
devreal: Typo?

```suggestion
inline MPI_Datatype mpi_type<unsigned long long>() {
```

devreal (commented):
Missing:

- `short` and `unsigned short`
- `char`, `unsigned char`, and `std::byte`

devreal (dismissed):

dutkalex:
TIL `std::size_t` is its own implementation-defined type, but `std::int8_t` etc are not apparently...

cwpearson:
@dutkalex if it's easier than SFINAE for this, I'd accept an implementation that is basically

```c++
template <T>
MPI_Datatype mpi_type() {
if constexpr (std::is_same_v<T, char>) {return MPI_CHAR; }
else if constexpr (std::is_same_v<T, int16_t>) { return ...; }
...
}
```

I think you can just have a branch for each type, and even if two branches are for the same type on certain platforms we won't get any warnings.

dutkalex:
@cwpearson Yes, this is actually how I intuitively implemented this in our in house solution, which explains why I did not run into the problem at the time. I will fix this ;)

cwpearson (dismissed):

cwpearson (dismissed):

src/impl/KokkosComm_types.hpp
```diff
@@ -21,50 +21,88 @@
 #include <Kokkos_Core.hpp>
 
 namespace KokkosComm::Impl {
-template <typename Scalar>
+template <typename T>
 MPI_Datatype mpi_type() {
-  static_assert(std::is_void_v<Scalar>, "mpi_type not implemented");
-  return MPI_CHAR;  // unreachable
-};
+  if constexpr (std::is_same_v<T, std::byte>)
```
devreal, line 28: Is `T` guaranteed to always be an unqualified type? Should we decay `T` to strip away potential const/volatile? Otherwise the logic won't work for `mpi_type<const double>`.

```suggestion
  using scalar_type = std::decay_t<T>;
  if constexpr (std::is_same_v<scalar_type, std::byte>)
```

devreal: Also, I don't know the kokkos coding rules. Some projects require braces around *all* blocks. Maybe that's not an issue here.

dutkalex: > Is `T` guaranteed to always be an unqualified type? Should we decay `T` to strip away potential const/volatile? Otherwise the logic won't work for `mpi_type<const double>`.

This function is meant to be called with T being a `View::value_type`, so I guess the question is whether `View<const double*>::value_type` is `double` or `const double`... But adding such a safety net can't hurt I guess. Any particular reason for using `std::decay_t` over `std::remove_cv_t`?

dutkalex: > Also, I don't know the kokkos coding rules. Some projects require braces around _all_ blocks. Maybe that's not an issue here.

I don't know either. @cwpearson what do you prefer?

devreal: Yes. This code will work with `std::decay_t` but not with `std::remove_cv_t`:

```C
template<typename T>
void send(T&& val) {
  MPI_Send(&val, 1, mpi_type<T>(), ...);
}

void foo() {
  int i;
  send(i);
}
```

`T` in this case will be `int&`. That's probably not going to happen with `View::value_type` but it might be used for something else in the future.

masterleinad: > I don't know either. @cwpearson what do you prefer?

This is something that is easily enforced with `clang-tidfy` when we decide that we care, see https://clang.llvm.org/extra/clang-tidy/checks/readability/braces-around-statements.html.

dutkalex: @devreal Maybe `std::remove_cvref_t` instead then, but `std::decay_t` also turn `T[]` into `T*` (and function types into function pointers too) and I don't think that we want this. What do you think?

devreal: Pointers are caught by the last branch (they're obviously not `void` so we get a `static_assert`. Same for functions.

dutkalex: You have a point, it does not matter indeed. Let's go with `decay` then, it's less typing 😅

devreal (dismissed):

src/impl/KokkosComm_types.hpp
```diff
@@ -23,48 +23,88 @@
...
+    return MPI_UNSIGNED;
+
+  else if constexpr (std::is_same_v<T, long>)
+    return MPI_LONG;
+  else if constexpr (std::is_same_v<T, unsigned long>)
+    return MPI_UNSIGNED_LONG;
+
+  else if constexpr (std::is_same_v<T, long long>)
+    return MPI_LONG_LONG;
+  else if constexpr (std::is_same_v<T, unsigned long>)
```
masterleinad:
```suggestion
  else if constexpr (std::is_same_v<T, unsigned long long>)
```

dutkalex:
Anything more we should add before merging this?

src/impl/KokkosComm_types.hpp
dutkalex: Nice catch, thanks!

devreal (approved):

masterleinad (approved):

PR #54, opened 2024-05-10
mpi: 1D contiguous alltoall in Impl

cwpearson:
Just in `Impl::` for now

### DIFF
--- a/src/impl/KokkosComm_alltoall.hpp
+++ b/src/impl/KokkosComm_alltoall.hpp
@@ -1,0 +1,116 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#pragma once
+
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#pragma once
+
+#include <Kokkos_Core.hpp>
+
+#include "KokkosComm_pack_traits.hpp"
+#include "KokkosComm_traits.hpp"
+
+// impl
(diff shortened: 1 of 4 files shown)


cwpearson:
Needs documentation

aprokop:
Needs `MPI_IN_PLACE` version.

cwpearson:
Okay we have an in-place version too now

src/impl/KokkosComm_alltoall.hpp
```diff
@@ -0,0 +1,116 @@
...
+      ss << "alltoall sendCount * communicator size (" << sendCount << " * " << size
+         << ") is greater than send view size";
+      throw std::runtime_error(ss.str());
+    }
+    if (recvCount * size > RT::extent(rv, 0)) {
+      std::stringstream ss;
+      ss << "alltoall recvCount * communicator size (" << recvCount << " * " << size
+         << ") is greater than recv view size";
+      throw std::runtime_error(ss.str());
+    }
```
cedricchevalier19, line 77: Do we want to always run these tests? Perhaps they can be conditionally compiled?

unit_tests/test_alltoall.cpp
```diff
@@ -0,0 +1,91 @@
...
+      sv.extent(0), KOKKOS_LAMBDA(const int i) { sv(i) = rank + i; });
+
+  KokkosComm::Impl::alltoall(Kokkos::DefaultExecutionSpace(), sv, nContrib, rv, nContrib, MPI_COMM_WORLD);
+
+  int errs;
+  Kokkos::parallel_reduce(
+      rv.extent(0),
+      KOKKOS_LAMBDA(const int &i, int &lsum) {
+        const int src = i / nContrib;                      // who sent this data
+        const int j   = rank * nContrib + (i % nContrib);  // what index i was at the source
```
cedricchevalier19, line 55:
```suggestion
        const int j   = i - rank * nContrib;  // what index i was at the source
```

unit_tests/test_alltoall.cpp
```diff
@@ -0,0 +1,91 @@
...
+      rv.extent(0), KOKKOS_LAMBDA(const int i) { rv(i) = rank + i; });
+
+  KokkosComm::Impl::alltoall(Kokkos::DefaultExecutionSpace(), rv, nContrib, MPI_COMM_WORLD);
+
+  int errs;
+  Kokkos::parallel_reduce(
+      rv.extent(0),
+      KOKKOS_LAMBDA(const int &i, int &lsum) {
+        const int src = i / nContrib;                      // who sent this data
+        const int j   = rank * nContrib + (i % nContrib);  // what index i was at the source
```
cedricchevalier19, line 84:
```suggestion
        const int j   = i - rank * nContrib;  // what index i was at the source
```

cedricchevalier19 (dismissed):
Just a small suggestion for getting rid of `%` operator.

src/impl/KokkosComm_alltoall.hpp
cwpearson: Yeah I think it's a good idea to skip these in a release build.

unit_tests/test_alltoall.cpp
cwpearson: Consider `i=0, rank=1, nContrib=1`, previously yeilds `1`, your suggestion yields `-1`

src/impl/KokkosComm_alltoall.hpp
cwpearson: I want to build a KOKKOSCOMM_DEBUG definition into the `KokkosComm_config.hpp` file you will introduce. I will open an issue to track that.

cedricchevalier19 (approved):

PR #32, opened 2024-04-12
non-contiguous `Impl::irecv`

cwpearson:
Replacement for https://github.com/kokkos/kokkos-comm/pull/9
- [x] Wait for https://github.com/kokkos/kokkos-comm/pull/55

Add an interface to KokkosComm:Req to register a callable at wait(). Use that interface to invoke unpack operations (if needed) at wait(). Add packer_type to MpiArgs.

PR #55, opened 2024-05-13
mpi: contiguous Irecv wrapper

cwpearson:


src/impl/KokkosComm_irecv.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include "KokkosComm_traits.hpp"
+
+// impl
+#include "KokkosComm_include_mpi.hpp"
+
+namespace KokkosComm::Impl {
+
+// low-level API
+template <KokkosView RecvView>
+void irecv(RecvView &rv, int src, int tag, MPI_Comm comm, MPI_Request &req) {
```
cedricchevalier19, line 33: Why not using `KokkosComm::Req` to be coherent with `isend`?

Also `isend` returns the request and here we pass it as reference, this can be confusing.

unit_tests/test_isendirecv.cpp
```diff
@@ -0,0 +1,109 @@
...
+  if (size < 2) {
+    GTEST_SKIP() << "Requires >= 2 ranks (" << size << " provided)";
+  }
+
+  if (0 == rank) {
+    int dst = 1;
+    Kokkos::parallel_for(
+        a.extent(0), KOKKOS_LAMBDA(const int i) { a(i) = i; });
+    KokkosComm::Req req = KokkosComm::isend(Kokkos::DefaultExecutionSpace(), a, dst, 0, MPI_COMM_WORLD);
+    req.wait();
```
cedricchevalier19, line 53: We can do also a classical `send` as `req` is not really used.

unit_tests/test_isendirecv.cpp
```diff
@@ -0,0 +1,109 @@
...
+  } else if (1 == rank) {
+    int src = 0;
+    MPI_Request req;
+    KokkosComm::Impl::irecv(a, src, 0, MPI_COMM_WORLD, req);
+    MPI_Wait(&req, MPI_STATUS_IGNORE);
```
cedricchevalier19, line 58: Mixing `KokkosComm` and `MPI_Wait` looks unintuitive.

unit_tests/view_builder.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+template <typename T, int RANK>
+struct ViewBuilder;
+
+template <typename T>
+struct ViewBuilder<T, 1> {
+  static auto view(noncontig, const std::string &name, int e0) {
+    // this is C-style layout, i.e. v(0,0) is next to v(0,1)
+    Kokkos::View<T **, Kokkos::LayoutRight> v(name, e0, 2);
+    return Kokkos::subview(v, Kokkos::ALL, 1);  // take column 1
```
cedricchevalier19, line 32: We have to be careful, `Kokkos::subview` is not symmetric:
I think a subview of a `LayoutRight` is still a `LayoutRight` while a subview of a `LayoutLeft` is a `LayoutStride`.
Not sure it change anything for us, but Kokkos will not do exactly the same thing underneath.

cedricchevalier19 (commented):
I think the API should be improved to be more coherent with `isend`.

Personally, I find the `isend` one better.

src/impl/KokkosComm_irecv.hpp
cwpearson: Sorry, I should have written a better description of this PR. Here I am just trying to add the lowest-possible level wrapper. There will still be a higher-level one followed in PR https://github.com/kokkos/kokkos-comm/pull/32

unit_tests/test_isendirecv.cpp
cwpearson: I'll update the test in https://github.com/kokkos/kokkos-comm/pull/32

unit_tests/view_builder.hpp
cwpearson: I think it has to be `LayoutStride` if it's not contiguous (not sure).

It's `LayoutStride` in this case: https://godbolt.org/z/Th888zzhh

unit_tests/test_isendirecv.cpp
cwpearson: True, that would be a `test_sendirecv.cpp` which I could also write.

cedricchevalier19: I do not think it is necessary.

unit_tests/view_builder.hpp
cedricchevalier19: We add an issue with this in the past, but I do not think it is this relevant here.

I think `Kokkos::LayoutRight` can be strided, and that is an important difference with `mdspan`.

cedricchevalier19 (approved):

PR #59, opened 2024-05-13
mpi: barrier wrapper

cwpearson:
Add the lowest useful level of wrapper for MPI_Barrier.

### DIFF
--- a/src/impl/KokkosComm_barrier.hpp
+++ b/src/impl/KokkosComm_barrier.hpp
@@ -27,12 +27,16 @@
 
 namespace KokkosComm::Impl {
 
-void barrier(MPI_Comm comm) { MPI_Barrier(comm); }
-
-template <KokkosExecutionSpace ExecSpace>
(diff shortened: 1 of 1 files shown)


src/impl/KokkosComm_barrier.hpp
```diff
@@ -26,6 +26,9 @@
 #include "KokkosComm_types.hpp"
 
 namespace KokkosComm::Impl {
+
+void barrier(MPI_Comm comm) { MPI_Barrier(comm); }
```
cedricchevalier19: Why no `Kokkos::Tools` here?

cwpearson: I was originally thinking we would really just put this inside a higher-level API and there would be a Kokkos Tools region in there, but on 2nd thought I think you're right, we can put regions in here too.

cedricchevalier19 (approved):

PR #68, opened 2024-05-23
Draft kokkos-comm initialization/finalization using `MPI_Session`s

dssgabriel:
This PR is a first attempt at initialization and finalization for `kokkos-comm` relying on `MPI_Session`s.

Some noteworthy additions brought by this PR:
- a `Communicator` class that wraps the `MPI_Comm` _and_ a Kokkos execution space
- a `Universe` class that holds the handle to the MPI sessions, as well as a session-associated communicator.

I expect lots of changes in the API, which I find kind of clunky at the moment. Reviews and comments on how to improve are welcome.

### DIFF
--- a/src/KokkosComm_communicator.hpp
+++ b/src/KokkosComm_communicator.hpp
@@ -18,6 +18,7 @@
 
 #include "KokkosComm_concepts.hpp"
 
+#include <Kokkos_Core_fwd.hpp>
 #include <cstdio>
 #include <mpi.h>
 
@@ -26,66 +27,37 @@ namespace KokkosComm {
 using Color = int;
 using Key   = int;
 
-template <KokkosExecutionSpace ExecSpace>
+template <KokkosExecutionSpace ExecSpace = Kokkos::DefaultExecutionSpace>
 class Communicator {
+ private:
+  MPI_Comm _comm;
+  ExecSpace _exec_space;
+
  public:
-  ~Communicator() {
-    switch (_comm_kind) {
-      // FIXME: find out how to properly free a session-associated communicator
-      // case CommunicatorKind::User: MPI_Comm_free(&_comm); break;
-      default: break;
-    }
-  }
-
-  static auto from_raw(MPI_Comm raw) -> Communicator {
-    assert(MPI_COMM_NULL != raw);
-
-    CommunicatorKind comm_kind;
-    if (MPI_COMM_SELF == raw) {
-      comm_kind = CommunicatorKind::Self;
-    } else if (MPI_COMM_WORLD == raw) {
-      comm_kind = CommunicatorKind::World;
-    } else {
-      int flag;
-      MPI_Comm_test_inter(raw, &flag);
-      if (0 == flag) {
-        comm_kind = CommunicatorKind::User;
-      } else {
-        fprintf(stderr, "[KokkosComm] error: intercommunicators are not supported (yet).\n");
-        std::terminate();
-      }
-    }
-
-    return Communicator(raw, comm_kind);
-  }
-
-  inline static auto from_raw_unchecked(MPI_Comm comm) -> Communicator {
-    return Communicator(comm, CommunicatorKind::User);
-  }
+  Communicator(MPI_Comm comm) : _comm(comm) {}
+  Communicator(const Communicator& other) = delete;
+  Communicator(const Communicator&& other) { _comm = std::move(other._comm); }
+  ~Communicator() { /*MPI_Comm_free(&_comm);*/ }
 
   static auto dup_raw(MPI_Comm raw) -> Communicator {
     MPI_Comm new_comm;
     MPI_Comm_dup(raw, &new_comm);
-    return Communicator(new_comm, CommunicatorKind::User);
+    return Communicator(new_comm);
   }
 
-  static auto dup(const Communicator &other) -> Communicator { return Communicator::dup_raw(other.as_raw()); }
+  static auto dup(const Communicator& other) -> Communicator { return Communicator::dup_raw(other.as_raw()); }
 
   static auto split_raw(MPI_Comm raw, Color color, Key key) -> Communicator {
     MPI_Comm new_comm;
     MPI_Comm_split(raw, color, key, &new_comm);
-    return Communicator(new_comm, CommunicatorKind::User);
+    return Communicator(new_comm);
   }
 
-  static auto split(const Communicator &other, Color color, Key key) -> Communicator {
+  static auto split(const Communicator& other, Color color, Key key) -> Communicator {
     return Communicator::split_raw(other.as_raw(), color, key);
   }
 
   inline auto as_raw() const -> MPI_Comm { return _comm; }
-
-  inline static auto self(void) -> Communicator { return Communicator::from_raw_unchecked(MPI_COMM_SELF); }
-
-  inline static auto world(void) -> Communicator { return Communicator::from_raw_unchecked(MPI_COMM_WORLD); }
 
   inline auto size(void) const -> int {
     int size;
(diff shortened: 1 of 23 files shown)


src/KokkosComm_communicator.hpp
```diff
@@ -0,0 +1,116 @@
...
+    } else if (MPI_COMM_WORLD == raw) {
+      comm_kind = CommunicatorKind::World;
+    } else {
+      int flag;
+      MPI_Comm_test_inter(raw, &flag);
+      if (0 == flag) {
+        comm_kind = CommunicatorKind::User;
+      } else {
+        fprintf(stderr, "[KokkosComm] error: intercommunicators are not supported (yet).\n");
+        std::terminate();
```
devreal: Don't just terminate. Either throw an exception or call `MPI_Abort` (I've seen applications hang where one process aborted without calling `MPI_Abort`, it's nasty)

src/KokkosComm_communicator.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+  static auto dup_raw(MPI_Comm raw) -> Communicator {
+    MPI_Comm new_comm;
+    MPI_Comm_dup(raw, &new_comm);
+    return Communicator(new_comm, CommunicatorKind::User);
+  }
+
+  static auto dup(const Communicator &other) -> Communicator { return Communicator::dup_raw(other.as_raw()); }
+
+  static auto split_raw(MPI_Comm raw, Color color, Key key) -> Communicator {
```
devreal, line 56: Is there a reason why this isn't just an overload of `split`?

src/KokkosComm_communicator.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+  static auto split(const Communicator &other, Color color, Key key) -> Communicator {
+    return Communicator::split_raw(other.as_raw(), color, key);
+  }
+
+  inline auto as_raw() const -> MPI_Comm { return _comm; }
+
+  inline static auto self(void) -> Communicator { return Communicator::from_raw_unchecked(MPI_COMM_SELF); }
+
+  inline static auto world(void) -> Communicator { return Communicator::from_raw_unchecked(MPI_COMM_WORLD); }
```
devreal: That doesn't belong into the communicator. Since we use sessions, there should be no use of the world process model. It could even be that no one has called `MPI_Init`.

src/impl/KokkosComm_MPI_instance.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+  // Resource release
+  MPI_Group_free(&kokkoscomm_group);
+  MPI_Info_free(&kokkoscomm_info);
+
+  return Universe<ExecSpace>(kokkoscomm_shandle, kokkoscomm_comm);
+}
+
+template <KokkosExecutionSpace ExecSpace>
+auto initialize(int &argc, char *argv[]) -> Universe<ExecSpace> {
```
devreal: It's not obvious that once version of `initialize` uses the WPM and one uses the SPM. I'd rather stay away from the WPM tbh.

src/KokkosComm_communicator.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include <cstdio>
+#include <mpi.h>
+
+namespace KokkosComm {
+
+using Color = int;
+using Key   = int;
+
+template <KokkosExecutionSpace ExecSpace>
+class Communicator {
```
devreal, line 31: What are the copy semantics of `Communicator`? I'd suggest to disable copying but allow moving. Some have suggested elsewhere that the copy constructor should call `MPI_Comm_dup` but makes implicit copy operations collective, which is not really intuitive.

devreal (changes requested):

src/KokkosComm_communicator.hpp
dssgabriel: Any specific error code you suggest?

src/KokkosComm_communicator.hpp
dssgabriel: I wanted to make it explicit that we're splitting from a raw `MPI_Comm` handle, not a wrapped `KokkosComm::Communicator`, but I guess we can also simply overload.

src/impl/KokkosComm_MPI_instance.hpp
dssgabriel: I'll remove the one using the WPM so that we only initialize using the SPM.

src/KokkosComm_communicator.hpp
dssgabriel: I will delete the copy constructor and only define a move one.

dssgabriel:
Thank you very much for the first review, @devreal!

Regarding C++ destruction and the SPM finalization semantics: is `MPI_Comm_free` required to be called before `MPI_Session_finalize`? And if so, how can I destroy the communicator without having to call its dtor explicitly?

src/KokkosComm_communicator.hpp
devreal: I guess that's a design decision that the Kokkos community should make (to be consistent with the other projects)

devreal:
Just a PSA: OMPI in combination with UCX won't support Sessions until the next major release of OMPI (6.0): https://github.com/open-mpi/ompi/issues/12566#issuecomment-2127642471

dssgabriel:
It looks like Linux's OpenMPI is too old to support `MPI_Session`s (which the initial support for was added in OpenMP 5.0), hence the failing CI tests.

@cwpearson Is there any way we can specifically install OpenMPI 5.x in the CI? The package from Ubuntu repos is simply not up-to-date.

src/KokkosComm_communicator.hpp
devreal: Not really, anything non-zero would work. This ties in with https://github.com/kokkos/kokkos-comm/discussions/29 so anything you do here is probably temporary anyway :)

devreal:
> Regarding C++ destruction and the SPM finalization semantics: is MPI_Comm_free required to be called before MPI_Session_finalize? And if so, how can I destroy the communicator without having to call its dtor explicitly?

The standard says that the application is required to clean up its objects before `MPI_Session_finalize`:

> The call to `MPI_SESSION_FINALIZE` does not free objects created by MPI calls; these
objects are freed using `MPI_XXX_FREE`, `MPI_COMM_DISCONNECT`, or
`MPI_FILE_CLOSE` calls.
Once `MPI_SESSION_FINALIZE` returns, no MPI procedure may be called in the Sessions
Model that are related to this session (not even freeing objects that are derived from
this session), except for those listed in Section 11.4.1.

We had a similar issue in other projects and came down to a registry in which objects that needed destruction were referenced. There are two options they can get destroyed:

1) Their destructor is called before the Session is destroyed. Then the object just removes itself from the registry.
2) They are released as part of the destruction of the session object (before `MPI_Session_finalize` is called) and the object becomes an empty shell, whose destructor does not do anything. 

All of this must be thread-safe etc but communicators are heavy objects that are not created and destroyed regularly so I think the overhead involved in such a scheme is acceptable.

src/KokkosComm_communicator.hpp
janciesko: You can drop the Raw. We typically are happy with overloads.

janciesko:
Would you like to rebase this on top of current develop?
@devreal, what's the status of sessions in some common MPI implementations? Can we drop this in or rather put it in an Experimental namespace?

devreal:
> @devreal, what's the status of sessions in some common MPI implementations? Can we drop this in or rather put it in an Experimental namespace?

Open MPI's session support has not been released yet. It will be included in the next major release (tbd when that comes out)

PR #69, opened 2024-06-03
MPI: in-place allgather for contiguous views

cwpearson:
Working towards building Arborx


cedricchevalier19 (approved):
Perhaps an `execution_space` version can make sense too.

PR #70, opened 2024-06-04
Move most things out of `Traits`

cwpearson:
Handling non-contiguous data can produce different View types, which makes using traits clumsy since they have to reference the declared type of the view. Just try removing them for now and see how we feel.

This also helps work towards building with CUDA

We can also revisit this if eventually we want to support non-view things.

### DIFF
--- a/src/impl/KokkosComm_isend.hpp
+++ b/src/impl/KokkosComm_isend.hpp
@@ -36,9 +36,9 @@ void isend(const SendView &sv, int dest, int tag, MPI_Comm comm, MPI_Request &re
   Kokkos::Tools::pushRegion("KokkosComm::Impl::isend");
   using KCT = typename KokkosComm::Traits<SendView>;
 
-  if (KCT::is_contiguous(sv)) {
+  if (KokkosComm::is_contiguous(sv)) {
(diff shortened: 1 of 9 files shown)


src/impl/KokkosComm_isend.hpp
```diff
@@ -77,13 +77,14 @@ KokkosComm::Req isend(const ExecSpace &space, const SendView &sv, int dest, int
 
     MpiArgs args = Packer::pack(space, sv);
     space.fence();
-    mpi_isend_fn(KCT::data_handle(args.view), args.count, args.datatype, dest, tag, comm, &req.mpi_req());
+    mpi_isend_fn(KokkosComm::data_handle(args.view), args.count, args.datatype, dest, tag, comm, &req.mpi_req());
```
cwpearson, line 80: This is an example of something that didn't work if the non-contiguous handler produced a different output view type in `args.view`.

cwpearson: I think ultimately the non-contiguous stuff in here will get re-written anyway

janciesko (approved):
LGTM

PR #72, opened 2024-06-04
Bump gtest to May 28,2024. Move unit test lambdas out to helper functions

### LINKED ISSUE #71: Can't use upstream gtest with nvcc
Since we use lambdas in the unit tests.

Produces errors like
```
.../unit_tests/test_alltoall.cpp(46): error: The enclosing parent function ("TestBody") for an extended __host__ __device__ lambda cannot have private or protected access within its class
                     __attribute__((host)) __attribute__((device))(const int i) { sv(i) = rank + i; });
                                                                  ^

.../unit_tests/test_alltoall.cpp(53): error: The enclosing parent function ("TestBody") for an extended __host__ __device__ lambda cannot have private or protected access within its class
       __attribute__((host)) __attribute__((device))(const int &i, int &lsum) {
                                                    ^

.../unit_tests/test_alltoall.cpp(75): error: The enclosing parent function ("TestBody") for an extended __host__ __device__ lambda cannot have private or protected access within its class
                     __attribute__((host)) __attribute__((device))(const int i) { rv(i) = rank + i; });
                                                                  ^

.../unit_tests/test_alltoall.cpp(82): error: The enclosing parent function ("TestBody") for an extended __host__ __device__ lambda cannot have private or protected access within its class
       __attribute__((host)) __attribute__((device))(const int &i, int &lsum) {
```

cwpearson:
Google Test does support lambdas in the test body with nvcc.

Fixes https://github.com/kokkos/kokkos-comm/issues/71


masterleinad (dismissed):
It's worth giving it a shot. 🙂

cwpearson:
Sorry @masterleinad I noticed a dumb mistake

dalg24 (commented):
That's a BAD idea
It means we will only be able to test against a patched googletest
Avoiding having kokkos lambdas directly in the body of the unit test and moving it to a helper function is not too bad of a constraint

cwpearson:
Okay @dalg24 with Daniel's help I figured out what you were getting at 👍

aprokop (approved):
LGTM

masterleinad (approved):

PR #75, opened 2024-06-05
Add CUDA + OpenMPI compile-only CI test

### LINKED ISSUE #73: Need a compile-only CUDA CI test
...then we can tackle getting some CI going at Sandia or elsewhere for actually running unit tests

cwpearson:
Prevent build regressions with nvcc

Needs https://github.com/kokkos/kokkos-comm/pull/72 before it will pass

Fixes https://github.com/kokkos/kokkos-comm/issues/73

### DIFF
--- a/.github/workflows/linux-compileonly.yaml
+++ b/.github/workflows/linux-compileonly.yaml
@@ -1,5 +1,11 @@
 name: Linux-CompileOnly
-on: push
+on:
+  pull_request:
+    paths-ignore:
+    - README.md
+    - 'docs/**'
+    types: [ opened, reopened, synchronize ]
+  workflow_dispatch:
 
 jobs:
   Kokkos-master-OpenMPI416-CUDA1250-Release:
--- a/.github/workflows/linux-compileonly.yaml
+++ b/.github/workflows/linux-compileonly.yaml
@@ -1,0 +1,58 @@
+name: Linux-CompileOnly
+on: push
+
+jobs:
+  Kokkos-master-OpenMPI416-CUDA1250-Release:
+    runs-on: ubuntu-latest
+    container:
+      image: nvidia/cuda:12.5.0-devel-ubuntu22.04
+      env:
+        KOKKOS_SRC: ${{ github.workspace }}/_deps/kokkos
+        KOKKOS_BUILD: ${{ github.workspace }}/_deps/kokkos-build
+        KOKKOS_INSTALL: ${{ github.workspace }}/_deps/kokkos-install
+        COMM_SRC: ${{ github.workspace }}
+        COMM_BUILD: ${{ github.workspace }}/build
+        timeout-minutes: 20
+    steps:
+      - name: Create workspace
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-compileonly.yaml
```diff
@@ -0,0 +1,64 @@
...
+          make -C openmpi-4.1.6 all install
+          ldconfig
+      - name: Print some MPI Info
+        run: |
+          which mpirun
+          mpirun --version
+          which mpicxx
+          mpicxx --version
+      - name: Check MPI CUDA Support
+        run: ompi_info --parsable --all | grep mpi_built_with_cuda_support:value:true
```
cedricchevalier19, line 55: It may make more sense to build a custom container image.

.github/workflows/linux-compileonly.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Build Kokkos
+        run: |
+          git clone https://github.com/kokkos/kokkos.git --branch master --depth 1 "$KOKKOS_SRC"
```
cedricchevalier19, line 58: I think getting explicitly the latest release is better. If so, and if we build our custom container image for tests, we can already put Kokkos inside.

cedricchevalier19 (approved):
I am firmly in favor of this feature. However, we should think about moving to a container to speed up the process and ease the development.

.github/workflows/linux-compileonly.yaml
```diff
@@ -0,0 +1,64 @@
...
+          cmake -S "$KOKKOS_SRC" -B "$KOKKOS_BUILD" -DCMAKE_CXX_COMPILER="$KOKKOS_SRC/bin/nvcc_wrapper" -DCMAKE_INSTALL_PREFIX="$KOKKOS_INSTALL" -DCMAKE_BUILD_TYPE=Release -DKokkos_ENABLE_SERIAL=ON -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_AMPERE80=ON
```
aprokop, line 59: Can we guarantee where the runner is going to run? Does it have A100?

aprokop (commented):
I agree with @cedricchevalier19 in creating a custom container. Building openmpi is not cheap.

.github/workflows/linux-compileonly.yaml
masterleinad: It's compile-only.

cwpearson: I figured I can get this up quicker than I can navigate the process of getting some CI going somewhere at Sandia, so let's just start with this for now.

cwpearson:
Ultimately the objective will be to actually run the unit tests at Sandia, which will also involve creating a container to run the tests in. It will take a bit to navigate that process, so we can do this in the mean time.

aprokop (approved):
4 min for OpenMPI is ok.

Issue #53, opened 2024-05-10
Implement MPI functions for ArborX


cwpearson:
@aprokop

- [x] MPI_Alltoall
- [x] MPI_Barrier
- [x] MPI_Allgather
  - [ ] in-place 
- [x] MPI_Irecv
- [x] MPI_Isend
- [ ] MPI Derived Datatypes

probably skip these for a bare MPI wrapper
- MPI_Comm_dup / MPI_Comm_free
- MPI_Init
- MPI_Finalize

aprokop:
This sounds right. Would be nice to also do something about getting rank and size from a communicator, but that's not a big deal.

Another pretty annoying thing that we have to do is to avoid passing `--help` to non-zero ranks, as otherwise Kokkos would print help on all ranks. We do it like [this](
https://github.com/arborx/ArborX/blob/d2f80a5d198ecace6ca6d123604dc5b0fb1afaeb/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp#L301-L315):
```c++
  // Strip "--help" and "--kokkos-help" from the flags passed to Kokkos if we
  // are not on MPI rank 0 to prevent Kokkos from printing the help message
  // multiply.
  if (comm_rank != 0)
  {
    auto *help_it = std::find_if(argv, argv + argc, [](std::string const &x) {
      return x == "--help" || x == "--kokkos-help";
    });
    if (help_it != argv + argc)
    {
      std::swap(*help_it, *(argv + argc - 1));
      --argc;
    }
  }
  Kokkos::initialize(argc, argv);
```

aprokop:
Another thing that could be worth trying to address at some point is to be able to run with GPU-aware MPI implementations, rather than doing everything on the host. We have a pathway in ArborX for it (even though we never saw it running faster, but it was like 4 years ago, maybe things are better now).

cwpearson:
I expect GPU-aware MPI to be pretty high on the priority list once we get some of these initial functions implemented!

janciesko:
GPU-aware MPI installations should be supported out of the box. Note that the communication itself is still host-triggered.

dutkalex:
> GPU-aware MPI installations should be supported out of the box. Note that the communication itself is still host-triggered.

[...]

PR #76, opened 2024-06-05
Add irecv wrapper in KokkosComm namespace

nicoleavans:
Currently, irecv is only available in the Impl namespace.

### DIFF
--- a/src/KokkosComm.hpp
+++ b/src/KokkosComm.hpp
@@ -22,6 +22,7 @@
 #include "KokkosComm_irecv.hpp"
 #include "KokkosComm_recv.hpp"
 #include "KokkosComm_send.hpp"
+#include "KokkosComm_alltoall.hpp"
 #include "KokkosComm_concepts.hpp"
 #include "KokkosComm_comm_mode.hpp"
 
@@ -35,8 +36,8 @@ Req isend(const ExecSpace &space, const SendView &sv, int dest, int tag, MPI_Com
 }
 
 template <KokkosView RecvView>
-void irecv(RecvView &sv, int src, int tag, MPI_Comm comm, MPI_Request req) {
-  return Impl::irecv(sv, src, tag, comm, req);
+void irecv(RecvView &rv, int src, int tag, MPI_Comm comm, MPI_Request req) {
+  return Impl::irecv(rv, src, tag, comm, req);
 }
 
 template <CommMode SendMode = CommMode::Default, KokkosExecutionSpace ExecSpace, KokkosView SendView>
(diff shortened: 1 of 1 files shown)


dssgabriel (dismissed):

nicoleavans:
Upon further thought, a function like this could be implemented to more evenly match up with the existing wrappers:

```
template <KokkosView RecvView>
void irecv(const ExecSpace &space, RecvView &sv, int src, int tag, MPI_Comm comm) {
  return Impl::irecv(space, sv, src, tag, comm);
}
```

aprokop:
I'd like this too for #53.

aprokop (approved):

aprokop:
Also needs `alltoall`

src/KokkosComm.hpp
```diff
@@ -33,6 +34,11 @@ Req isend(const ExecSpace &space, const SendView &sv, int dest, int tag, MPI_Com
   return Impl::isend<SendMode>(space, sv, dest, tag, comm);
 }
 
+template <KokkosView RecvView>
+void irecv(RecvView &sv, int src, int tag, MPI_Comm comm, MPI_Request req) {
```
dssgabriel: Why use `sv` for the receiving view? I think it was previously clearer with `rv` instead.

dssgabriel:
As the `MPI_Alltoall` wrapper has been implemented, consider adding it to this PR as well

src/KokkosComm.hpp
nicoleavans: I was trying to more closely match the `recv` wrapper in the file:
```
void recv(const ExecSpace &space, RecvView &sv, int src, int tag, MPI_Comm comm) {
  return Impl::recv(space, sv, src, tag, comm);
}
```

I agree it is clearer, I can change it back.

dssgabriel: I think you can change the `recv` wrapper to `RecvView &rv` as well then :+1:

src/KokkosComm.hpp
```diff
@@ -33,14 +35,25 @@ Req isend(const ExecSpace &space, const SendView &sv, int dest, int tag, MPI_Com
...
+
 template <CommMode SendMode = CommMode::Default, KokkosExecutionSpace ExecSpace, KokkosView SendView>
 void send(const ExecSpace &space, const SendView &sv, int dest, int tag, MPI_Comm comm) {
   return Impl::send<SendMode>(space, sv, dest, tag, comm);
 }
 
 template <KokkosExecutionSpace ExecSpace, KokkosView RecvView>
-void recv(const ExecSpace &space, RecvView &sv, int src, int tag, MPI_Comm comm) {
-  return Impl::recv(space, sv, src, tag, comm);
+void recv(const ExecSpace &space, RecvView &rv, int src, int tag, MPI_Comm comm) {
```
cwpearson, line 49: 👍

cwpearson (approved):

aprokop (commented):
If all we do right now is just passing things through, can't we simply do
```c++
  namespace KokkosComm {
    using Impl::recv;
    using Impl::alltoall;
  }
```
Edit: was too late with the review.

PR #81, opened 2024-06-11
Use MPI_BYTE for unknown types

### LINKED ISSUE #79: Convert user types to `MPI_BYTE`
This logic seems problematic to me:

https://github.com/kokkos/kokkos-comm/blob/fc25da1fd67a325a47ad9d6584d00a82aa74fa8b/src/impl/KokkosComm_types.hpp#L102-L105

If I have a `View<Box*>`, with `Box` being some POD struct, in my opinion that view should be sent with message size `view.size() * sizeof(View::value_type)` and MPI type `MPI_BYTE`. Instead, right now it refuses to deal with it.

aprokop:
Fix #79.

### DIFF
--- a/src/impl/KokkosComm_types.hpp
+++ b/src/impl/KokkosComm_types.hpp
@@ -25,10 +25,7 @@ template <typename Scalar>
 MPI_Datatype mpi_type() {
   using T = std::decay_t<Scalar>;
 
-  if constexpr (std::is_same_v<T, std::byte>)
-    return MPI_BYTE;
-
-  else if constexpr (std::is_same_v<T, char>)
+  if constexpr (std::is_same_v<T, char>)
     return MPI_CHAR;
   else if constexpr (std::is_same_v<T, unsigned char>)
     return MPI_UNSIGNED_CHAR;
@@ -98,6 +95,9 @@ MPI_Datatype mpi_type() {
     return MPI_COMPLEX;
   else if constexpr (std::is_same_v<T, Kokkos::complex<double>>)
     return MPI_DOUBLE_COMPLEX;
+
+  else if constexpr (std::is_trivially_copyable_v<T>)
+    return MPI_BYTE;
 
   else {
     static_assert(std::is_void_v<T>, "mpi_type not implemented");


src/impl/KokkosComm_types.hpp
```diff
@@ -99,10 +96,8 @@ MPI_Datatype mpi_type() {
   else if constexpr (std::is_same_v<T, Kokkos::complex<double>>)
     return MPI_DOUBLE_COMPLEX;
 
-  else {
-    static_assert(std::is_void_v<T>, "mpi_type not implemented");
```
masterleinad, line 103: It might make sense to assert that `T` is trivially copyable at least.

aprokop: Good point.

masterleinad:
To actually fix #79 you would also need to make sure that you scale the number of bytes to send and receive accordingly, though.

aprokop:
> To actually fix #79 you would also need to make sure that you scale the number of bytes to send and receive accordingly, though.

I feel sheepish for forgetting this.

src/impl/KokkosComm_types.hpp
cwpearson: Looking ahead, I think trivially-copyable is a problem for Trilinos FAD types - they are actually trivially-copyable, but because of how they are implemented they have non-default ctors, etc that makes C++ think they're not trivially-copyable.

cwpearson: Maybe if we run into this problem (a problem of success), we can make a special configure-time toggle to remove this trivially-copyable check.

aprokop: One thing I'm not a fan of is that if you have type T with large `sizeof(T)`, `n * sizeof(T)` may exceed `INT_MAX`, while storing them in a view is totally fine.

cwpearson: For now, maybe we can just send as `MPI_BYTE` as long as the size (MPI_Type_size) matches the extent (MPI_Type_get_extent) and the lower bound (MPI_Type_get_extent) is 0; is this the best way to check if a type is contiguous?

src/impl/KokkosComm_types.hpp
```diff
@@ -99,6 +96,9 @@ MPI_Datatype mpi_type() {
   else if constexpr (std::is_same_v<T, Kokkos::complex<double>>)
     return MPI_DOUBLE_COMPLEX;
 
+  else if constexpr (std::is_trivially_copyable_v<T>)
+    return MPI_BYTE;
```
devreal, line 100: I'm not sure that this fix alone solves your problem. I'm missing the part where we get the size of T. In https://github.com/kokkos/kokkos-comm/blob/develop/src/impl/KokkosComm_send.hpp#L72 we only get the type (`MPI_BYTE` here) and take the span, not the `span*sizeof(T)`. Maybe `mpi_type()` should provide a count multiplier as well?

src/impl/KokkosComm_types.hpp
devreal: Kokkos could offer a trait that overrides `std::is_trivially_copyable` so that users can tell Kokkos that they know what they are doing (e.g., not trying to copy pointers between processes).

src/impl/KokkosComm_types.hpp
aprokop: Yes, @masterleinad identified this problem. I'm not sure what's the best way to proceed here. Your suggestion is one of the possible variants.

src/impl/KokkosComm_types.hpp
masterleinad: We discussed that also in the context of `deep_copy` where we are assuming that all types used in kernels are logically trivially copyable. There, we decided that this would be too big of a change to error out unless the user has specialized such a trait.

cwpearson:
What if the fallback case was just to construct, register, and cache a contiguous MPI derived datatype that is sizeof T in a static variable.

aprokop:
> What if the fallback case was just to construct, register, and cache a contiguous MPI derived datatype that is sizeof T in a static variable.

That may work. I'm not familiar with it. Could you post a snippet of code to demonstrate what you are thinking?

cwpearson:
Untested:

```c++
else /*constexpr*/ {
  static MPI_Datatype cached = MPI_DATATYPE_NULL;
  if (MPI_DATATYPE_NULL == cached) {
    MPI_Type_contiguous(sizeof(T), MPI_BYTE, &cached);
    MPI_Type_commit(&cached);
  }
  return cached;
}
```
each call to mpi_type<T> with a new T that reaches the fallback with a different T will have a different `cached` variable. Each new `T` will be created and committed once, lazily, the first time it is seen. Otherwise it will be a quick (?) runtime check.

aprokop:
> each call to mpi_type with a new T that reaches the fallback with a different T will have a different `cached` variable

Oh, that's a cool idea! This sounds like it could work well. Maybe create a templated function to create mpi types, so that users could specialize it on their own. With the default being the two MPI calls you posted.

cwpearson:
Yeah it would be great if this could be exposed to the user to override it. I think we'd have to refactor this to be a SFINAE-style implementation rather than `if constexpr` then?

aprokop:
> Yeah it would be great if this could be exposed to the user to override it. I think we'd have to refactor this to be a SFINAE-style implementation rather than `if constexpr` then?

Yeah, probably just a templated function with the default type creation and specializations.

devreal:
Who frees the temporary type? It should probably be put into a list of types that get destroyed at the end...

aprokop:
~~Can make the static variable a holder class that frees on destruction.~~ Nevermind, it would be called too late, after MPI was finalized.

cwpearson:
Yeah each view, not each view type, will have a separate MPI_Datatype...harder to cache for sure. I was thinking too quickly.

cwpearson:
```c++
template <KokkosView View>
MPI_Datatype view_mpi_type(const View &view) {

  constexpr int P = 2*View::rank;
  using Key = std::array<int, P>;

  auto key_from = [](const View &v) -> Key {
    Key key;
    for (size_t d = 0; d < View::rank; d++) {
      key[d] = v.extent(d);
      key[d+1] = v.stride(d);
    }
    return key;
  };


  static std::map<Key, MPI_Datatype> cache;

  Key key = key_from(view);
  if (cache.count(key) > 0) {
    return cache[key];
  }


  using value_type = typename View::non_const_value_type;
  MPI_Datatype type = mpi_type_v<value_type>;

  // This doesn't work for 1D contiguous views into reduce because it
  // represents the whole 1D view as 1 Hvector, rather than N elements.
  // FIXME: is there a more generic way to handle this, maybe by treating
  // the last dimension specially under certain circumstances?
  for (size_t d = 0; d < KokkosComm::rank<View>(); ++d) {
    MPI_Datatype newtype;
    MPI_Type_create_hvector(KokkosComm::extent(view, d) /*count*/, 1 /*block length*/,
                            KokkosComm::stride(view, d) * sizeof(value_type), type, &newtype);
    type = newtype;
  }
  MPI_Type_commit(&type);
  cache[key] = type;
  return type;
}
```

cwpearson:
Are we actually required to call MPI_Type_free?

aprokop:
> Are we actually required to call MPI_Type_free?

Some (like [this](https://hpc-forge.cineca.it/files/ScuolaCalcoloParallelo_WebDAV/public/anno-2016/25_Summer_School/Bologna/MPI_Derived_Datatypes.pdf) presentation, slide 4) state that "You are not required to use it or deallocate it, but it is recommended (there may be a limit)". 

I didn't see anything in the standard saying that it is required.

devreal:
It's not required. Datatypes are a funny thing in MPI because they can persist between Sessions (mostly for historically reasons, because they are not bound to any Session). Having said that, KokkosComm should strive to exit cleanly with all resources freed. This makes it easier to use leak detection tools and is just good practice.

cwpearson:
This ties in to the init / finalize discussions I think. We have to have the opportunity to do something before / when MPI_Finalize gets called.

devreal:
We could have a function that returns a  `std:pair<MPI_Datatype,
std::size_t>`. For basic types `.second` is 1, for derived types it's
`sizeof(T)` and `.first` is `MPI_BYTE`. It's all `constexpr` and with big
count in MPI we don't need to worry about integer overflows.

PR #84, opened 2024-06-12
Add fences for alltoall

cedricchevalier19:
`alltoall(Execspace, ...` were missing fences


masterleinad (approved):

aprokop (approved):
LGTM.

Currently it is a bit weird, as we don't even require that the view's memory space is accessible from the execution space. Another thing is that the we should probably change most (all?) fences to also have the label.

cedricchevalier19:
> Currently it is a bit weird, as we don't even require that the view's memory space is accessible from the execution space. 

It's a very good point.

> Another thing is that the we should probably change most (all?) fences to also have the label.

I have copied one with labels, but we should update all the others. Thanks for noticing.

PR #88, opened 2024-06-14
Provide `KokkosComm::initialize`/`finalize`

### LINKED ISSUE #82: Provide a KokkosCommInitialize
It must enforce a correct calling order for `MPI_Init` and `Kokkos::initialize`.

It should also handle special Kokkos arguments:

> Another pretty annoying thing that we have to do is to avoid passing `--help` to non-zero ranks, as otherwise Kokkos would print help on all ranks. We do it like [this](
https://github.com/arborx/ArborX/blob/d2f80a5d198ecace6ca6d123604dc5b0fb1afaeb/benchmarks/distributed_tree_driver/distributed_tree_driver.cpp#L301-L315):
```c++
  // Strip "--help" and "--kokkos-help" from the flags passed to Kokkos if we
  // are not on MPI rank 0 to prevent Kokkos from printing the help message
  // multiply.
  if (comm_rank != 0)
  {
    auto *help_it = std::find_if(argv, argv + argc, [](std::string const &x) {
      return x == "--help" || x == "--kokkos-help";
    });
    if (help_it != argv + argc)
    {
      std::swap(*help_it, *(argv + argc - 1));
      --argc;
    }
  }
  Kokkos::initialize(argc, argv);
```

_Originally posted by @aprokop in https://github.com/kokkos/kokkos-comm/issues/53#issuecomment-2105364199_

dssgabriel:
This PR implements a very basic initialization and finalization for KokkosComm (closes #82).

Users should now only call `KokkosComm::initialize` and `KokkosComm::finalize`, no more manual MPI and Kokkos initialization. Using these functions ensures MPI is correctly initialized/finalized (with multiple thread support) _before_ Kokkos is initialized/finalized.

To-do before merging:
- [x] add documentation on these functions
- [ ] make it clear to users that they should only call `KokkosComm` when initializing/finalizing a Kokkos + MPI application

### DIFF
--- a/src/KokkosComm.hpp
+++ b/src/KokkosComm.hpp
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include "KokkosComm_include_mpi.hpp"
 #include "KokkosComm_collective.hpp"
 #include "KokkosComm_version.hpp"
 #include "KokkosComm_isend.hpp"
@@ -28,7 +29,55 @@
 
 #include <Kokkos_Core.hpp>
 
+#include <algorithm>
+#include <cstdio>
+#include <string_view>
+
 namespace KokkosComm {
+
+inline void initialize(int &argc, char ***argv) {
+  int flag;
+  MPI_Initialized(&flag);
+  // Eagerly abort if MPI has already been initialized
+  if (0 != flag) {
+    int rank;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+    if (0 == rank) {
+      fprintf(stderr, "error: MPI must not be initialized prior to initializing KokkosComm\n");
+    }
+    MPI_Abort(MPI_COMM_WORLD, -1);
+  }
+
+  int provided;
+  MPI_Init_thread(&argc, argv, MPI_THREAD_MULTIPLE, &provided);
+  int rank;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+  // Abort if MPI failed to provide Thread Multiple
+  if (MPI_THREAD_MULTIPLE != provided) {
+    if (0 == rank) {
+      fprintf(stderr, "error: failed to initialized with required thread support\n");
+    }
+    MPI_Abort(MPI_COMM_WORLD, -1);
+  }
+
+  // Strip "--help" and "--kokkos-help" from the flags passed to Kokkos if we are not on rank 0 to prevent Kokkos
+  // from printing the help message multiple times.
+  if (0 != rank) {
+    auto *help_it = std::find_if(*argv, *argv + argc,
+                                 [](std::string_view const &x) { return x == "--help" || x == "--kokkos-help"; });
+    if (help_it != *argv + argc) {
+      std::swap(*help_it, *(*argv + argc - 1));
+      --argc;
+    }
+  }
+  Kokkos::initialize(argc, *argv);
+}
+
+inline void finalize() {
+  Kokkos::finalize();
+  MPI_Finalize();
+}
 
 template <CommMode SendMode = CommMode::Default, KokkosExecutionSpace ExecSpace, KokkosView SendView>
 Req isend(const ExecSpace &space, const SendView &sv, int dest, int tag, MPI_Comm comm) {
--- a/src/KokkosComm.hpp
+++ b/src/KokkosComm.hpp
@@ -40,27 +40,18 @@ inline void initialize(int &argc, char ***argv) {
   MPI_Initialized(&flag);
   // Eagerly abort if MPI has already been initialized
   if (0 != flag) {
-    int rank;
-    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
-    if (0 == rank) {
-      fprintf(stderr, "error: MPI must not be initialized prior to initializing KokkosComm\n");
-    }
     MPI_Abort(MPI_COMM_WORLD, -1);
   }
 
   int provided;
   MPI_Init_thread(&argc, argv, MPI_THREAD_MULTIPLE, &provided);
-  int rank;
-  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
-
   // Abort if MPI failed to provide Thread Multiple
   if (MPI_THREAD_MULTIPLE != provided) {
-    if (0 == rank) {
-      fprintf(stderr, "error: failed to initialized with required thread support\n");
-    }
     MPI_Abort(MPI_COMM_WORLD, -1);
   }
 
+  int rank;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
   // Strip "--help" and "--kokkos-help" from the flags passed to Kokkos if we are not on rank 0 to prevent Kokkos
   // from printing the help message multiple times.
   if (0 != rank) {
--- a/src/KokkosComm.hpp
+++ b/src/KokkosComm.hpp
@@ -35,7 +35,7 @@
 
 namespace KokkosComm {
 
-inline void initialize(int &argc, char ***argv) {
+inline void initialize(int &argc, char *argv[], int mpi_required_thread_lvl) {
   int flag;
   MPI_Initialized(&flag);
   // Eagerly abort if MPI has already been initialized
@@ -44,9 +44,9 @@ inline void initialize(int &argc, char ***argv) {
   }
 
   int provided;
-  MPI_Init_thread(&argc, argv, MPI_THREAD_MULTIPLE, &provided);
-  // Abort if MPI failed to provide Thread Multiple
-  if (MPI_THREAD_MULTIPLE != provided) {
+  MPI_Init_thread(&argc, &argv, mpi_required_thread_lvl, &provided);
+  // Abort if MPI failed to provide the required thread level
+  if (mpi_required_thread_lvl != provided) {
     MPI_Abort(MPI_COMM_WORLD, -1);
   }
 
@@ -55,14 +55,18 @@ inline void initialize(int &argc, char ***argv) {
   // Strip "--help" and "--kokkos-help" from the flags passed to Kokkos if we are not on rank 0 to prevent Kokkos
   // from printing the help message multiple times.
   if (0 != rank) {
-    auto *help_it = std::find_if(*argv, *argv + argc,
+    auto *help_it = std::find_if(argv, argv + argc,
                                  [](std::string_view const &x) { return x == "--help" || x == "--kokkos-help"; });
-    if (help_it != *argv + argc) {
-      std::swap(*help_it, *(*argv + argc - 1));
+    if (help_it != argv + argc) {
+      std::swap(*help_it, *(argv + argc - 1));
       --argc;
     }
   }
-  Kokkos::initialize(argc, *argv);
+  Kokkos::initialize(argc, argv);
+}
+
+inline void initialize(int &argc, char *argv[]) {
+  initialize(argc, argv, MPI_THREAD_MULTIPLE);
 }
 
 inline void finalize() {
--- a/src/KokkosComm.hpp
+++ b/src/KokkosComm.hpp
@@ -45,9 +45,14 @@ enum class ThreadSupportLevel {
 inline void initialize(int &argc, char *argv[], ThreadSupportLevel required) {
   int flag;
   MPI_Initialized(&flag);
-  // Eagerly abort if MPI has already been initialized
+  // Forbid calling this function if MPI has already been initialized
   if (0 != flag) {
     MPI_Abort(MPI_COMM_WORLD, -1);
+  }
+
+  // Forbid calling this function if Kokkos has already been initialized
+  if (Kokkos::is_initialized()) {
+    std::abort();
   }
 
   int provided;
@@ -74,7 +79,18 @@ inline void initialize(int &argc, char *argv[], ThreadSupportLevel required) {
 inline void initialize(int &argc, char *argv[]) { initialize(argc, argv, ThreadSupportLevel::Multiple); }
 
 inline void finalize() {
+  // Forbid calling this function if Kokkos has already been finalized or isn't yet initialized
+  if (Kokkos::is_finalized() || !Kokkos::is_initialized()) {
+    MPI_Abort(MPI_COMM_WORLD, -1);
+  }
   Kokkos::finalize();
+
+  int flag;
+  MPI_Finalized(&flag);
+  // Forbid calling this function if MPI has already been finalized
+  if (0 != flag) {
+    std::abort();
+  }
   MPI_Finalize();
 }
 
--- a/src/KokkosComm.hpp
+++ b/src/KokkosComm.hpp
@@ -42,7 +42,7 @@ enum class ThreadSupportLevel {
   Multiple   = MPI_THREAD_MULTIPLE,
 };
 
-inline void initialize(int &argc, char *argv[], ThreadSupportLevel required) {
+inline void initialize(int &argc, char *argv[], ThreadSupportLevel required = ThreadSupportLevel::Multiple) {
   int flag;
   MPI_Initialized(&flag);
   // Forbid calling this function if MPI has already been initialized
@@ -75,8 +75,6 @@ inline void initialize(int &argc, char *argv[], ThreadSupportLevel required) {
   }
   Kokkos::initialize(argc, argv);
 }
-
-inline void initialize(int &argc, char *argv[]) { initialize(argc, argv, ThreadSupportLevel::Multiple); }
 
 inline void finalize() {
   // Forbid calling this function if Kokkos has already been finalized or isn't yet initialized
--- a/src/KokkosComm.hpp
+++ b/src/KokkosComm.hpp
@@ -35,7 +35,14 @@
 
 namespace KokkosComm {
 
-inline void initialize(int &argc, char *argv[], int mpi_required_thread_lvl) {
+enum class ThreadSupportLevel {
+  Single     = MPI_THREAD_SINGLE,
+  Funneled   = MPI_THREAD_FUNNELED,
+  Serialized = MPI_THREAD_SERIALIZED,
+  Multiple   = MPI_THREAD_MULTIPLE,
+};
+
+inline void initialize(int &argc, char *argv[], ThreadSupportLevel required) {
   int flag;
   MPI_Initialized(&flag);
   // Eagerly abort if MPI has already been initialized
@@ -44,9 +51,9 @@ inline void initialize(int &argc, char *argv[], int mpi_required_thread_lvl) {
   }
 
   int provided;
-  MPI_Init_thread(&argc, &argv, mpi_required_thread_lvl, &provided);
-  // Abort if MPI failed to provide the required thread level
-  if (mpi_required_thread_lvl != provided) {
+  MPI_Init_thread(&argc, &argv, static_cast<int>(required), &provided);
+  // Abort if MPI failed to provide the required thread support level
+  if (static_cast<int>(required) < provided) {
     MPI_Abort(MPI_COMM_WORLD, -1);
   }
 
@@ -65,7 +72,7 @@ inline void initialize(int &argc, char *argv[], int mpi_required_thread_lvl) {
   Kokkos::initialize(argc, argv);
 }
 
-inline void initialize(int &argc, char *argv[]) { initialize(argc, argv, MPI_THREAD_MULTIPLE); }
+inline void initialize(int &argc, char *argv[]) { initialize(argc, argv, ThreadSupportLevel::Multiple); }
 
 inline void finalize() {
   Kokkos::finalize();
--- a/src/KokkosComm.hpp
+++ b/src/KokkosComm.hpp
@@ -62,9 +62,8 @@ inline void initialize(int &argc, char *argv[], ThreadSupportLevel required) {
   // Strip "--help" and "--kokkos-help" from the flags passed to Kokkos if we are not on rank 0 to prevent Kokkos
   // from printing the help message multiple times.
   if (0 != rank) {
-    auto *help_it = std::find_if(
-        argv, argv + argc, [](std::string_view const &x) { return x.find("--kokkos-") != std::string_view::npos; });
-    if (help_it != argv + argc) {
+    if (auto *help_it = std::find_if(argv, argv + argc, [](std::string_view const &x) { return x == "--kokkos-help"; });
+        help_it != argv + argc) {
       std::swap(*help_it, *(argv + argc - 1));
       --argc;
     }
--- a/src/KokkosComm.hpp
+++ b/src/KokkosComm.hpp
@@ -55,8 +55,8 @@ inline void initialize(int &argc, char *argv[], int mpi_required_thread_lvl) {
   // Strip "--help" and "--kokkos-help" from the flags passed to Kokkos if we are not on rank 0 to prevent Kokkos
   // from printing the help message multiple times.
   if (0 != rank) {
-    auto *help_it = std::find_if(argv, argv + argc,
-                                 [](std::string_view const &x) { return x == "--help" || x == "--kokkos-help"; });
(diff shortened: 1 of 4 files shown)


src/KokkosComm.hpp
```diff
@@ -28,8 +29,56 @@
...
+  MPI_Initialized(&flag);
+  // Eagerly abort if MPI has already been initialized
+  if (0 != flag) {
+    int rank;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+    if (0 == rank) {
+      fprintf(stderr, "error: MPI must not be initialized prior to initializing KokkosComm\n");
+    }
+    MPI_Abort(MPI_COMM_WORLD, -1);
+  }
```
cedricchevalier19: I'm not sure if we need to abort in this case.

src/KokkosComm.hpp
```diff
@@ -28,8 +29,56 @@
...
+
+  int provided;
+  MPI_Init_thread(&argc, argv, MPI_THREAD_MULTIPLE, &provided);
```
cedricchevalier19: Do we always need `MPI_THREAD_MULTIPLE`?
On the Kokkos side, it is not required if all host parallel backends are disabled.

src/KokkosComm.hpp
```diff
@@ -28,8 +29,56 @@
...
+  int rank;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+  // Abort if MPI failed to provide Thread Multiple
+  if (MPI_THREAD_MULTIPLE != provided) {
+    if (0 == rank) {
+      fprintf(stderr, "error: failed to initialized with required thread support\n");
+    }
+    MPI_Abort(MPI_COMM_WORLD, -1);
+  }
```
cedricchevalier19: It is related to the previous comment.

src/KokkosComm.hpp
```diff
@@ -28,8 +29,56 @@
...
+  // Strip "--help" and "--kokkos-help" from the flags passed to Kokkos if we are not on rank 0 to prevent Kokkos
+  // from printing the help message multiple times.
+  if (0 != rank) {
+    auto *help_it = std::find_if(*argv, *argv + argc,
+                                 [](std::string_view const &x) { return x == "--help" || x == "--kokkos-help"; });
+    if (help_it != *argv + argc) {
+      std::swap(*help_it, *(*argv + argc - 1));
+      --argc;
+    }
+  }
```
cedricchevalier19: I am unsure if striping the `--help` is interesting. To me, it is not necessarily a Kokkos' option.

cedricchevalier19 (commented):
Some thoughts, and be precise in the documentation.

src/KokkosComm.hpp
dssgabriel: This is just to ensure that the application "correctly" terminates if MPI was already initialized (as @devreal had suggested in #68). I am not sure if that is really necessary here but I guess it is cleaner?

src/KokkosComm.hpp
dssgabriel: I am not sure :thinking: I believe MPI implementations eagerly enable `THREAD_MULTIPLE` support nowadays anyway. MPI experts should be able to tell us what to do here.

src/KokkosComm.hpp
dssgabriel: If `MPI_THREAD_MULTIPLE` is not _really_ needed, I guess this whole `if` statement can be deleted.

src/KokkosComm.hpp
dssgabriel: Sure, I'll remove it! I pretty much copy-pasted it from the issue :slightly_smiling_face:

aprokop: @cedricchevalier19 So you would want to print `--help` multiple times from every rank?

aprokop (commented):
I think it would be nice to have `KokkosComm::ScopeGuard` that is similar to `Kokkos::ScopeGuard` in addition to `finalize` and `initialize`. That way, a user won't have to remember to call `finalize`.

dssgabriel:
> I think it would be nice to have `KokkosComm::ScopeGuard` that is similar to `Kokkos::ScopeGuard`

Should this and other utilities (e.g. `is_initialized`) be added here or in another PR?

aprokop:
> Should this and other utilities (e.g. `is_initialized`) be added here or in another PR?

Easier in a different PR to keep the PR size small.

src/KokkosComm.hpp
```diff
@@ -28,8 +29,56 @@
 
 #include <Kokkos_Core.hpp>
 
+#include <algorithm>
+#include <cstdio>
+#include <string_view>
+
 namespace KokkosComm {
 
+inline void initialize(int &argc, char ***argv) {
```
aprokop: Do you really need `***`, or is `char *argv[]` sufficient?

src/KokkosComm.hpp
```diff
@@ -28,8 +29,56 @@
...
+  int flag;
+  MPI_Initialized(&flag);
+  // Eagerly abort if MPI has already been initialized
+  if (0 != flag) {
+    int rank;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+    if (0 == rank) {
+      fprintf(stderr, "error: MPI must not be initialized prior to initializing KokkosComm\n");
+    }
+    MPI_Abort(MPI_COMM_WORLD, -1);
```
aprokop: From a reading perspective, it's better to keep the `if` clauses smaller
```c++
if (flag != 0)
  MPI_Abort(MPI_COMM_WORLD, -1);

int rank;
...
```

src/KokkosComm.hpp
aprokop: I think this needs a semantic discussion. I don't feel like I fully understand the ramifications of this check.

src/KokkosComm.hpp
aprokop: Maybe it should be an optional argument to `initialize`. Would not want to take away control from users.

src/KokkosComm.hpp
dssgabriel: Should be sufficient if I pass it as `&argv` to `MPI_Init` :+1:

src/KokkosComm.hpp
cedricchevalier19: I would tend to protect all `--kokkos-*` options but let the users do whatever they want with their other arguments.

src/KokkosComm.hpp
dssgabriel: The idea is that users should not call MPI nor Kokkos initialization/finalization functions themselves (at least for now). This lets us ensure MPI is initialized _before_ Kokkos.

aprokop: I understand that part. What i don't quite understand is what it means if you call initialize function after MPI has already been initialized somewhere, when it happens, and whether it should be allowed.

If you allow MPI to be initialized prior to calling initialize, shouldn't you then not be allowed to finalize MPI in the finalize call?

aprokop: I'd rather always require MPI initialization in KokkosComm's initialize and finalize. If a user knows MPI initialization/finalization is done somewhere else, then just use Kokkos initialize/finalize explicitly.

I agree with the current code, and i think am against the suggestion to make it optional.

dssgabriel: I am not sure I understand what you mean here. Do you want to:

**(a)** forbid users to initialize/finalize MPI themselves, i.e. everything shall be done with KokkosComm's initialization and finalization functions; or
**(b)** allow users to initialize/finalize MPI themselves (then KokkosComm only initializes/finalizes Kokkos)?

I, and probably @cedricchevalier19 too, would prefer going with option (a) and enforcing it, at least for now.

src/KokkosComm.hpp
```diff
@@ -55,8 +55,8 @@ inline void initialize(int &argc, char *argv[], int mpi_required_thread_lvl) {
   // Strip "--help" and "--kokkos-help" from the flags passed to Kokkos if we are not on rank 0 to prevent Kokkos
   // from printing the help message multiple times.
   if (0 != rank) {
-    auto *help_it = std::find_if(argv, argv + argc,
-                                 [](std::string_view const &x) { return x == "--help" || x == "--kokkos-help"; });
+    auto *help_it = std::find_if(
+        argv, argv + argc, [](std::string_view const &x) { return x.find("--kokkos-") != std::string_view::npos; });
```
cedricchevalier19: In fact, after some thought, it is less clear. Some `--kokkos-*` options should still be passed to `Kokkos::initialize`. 

I wonder if the options to bind to the device should be treated by Kokkos or KokkosComm.

KokkosComm has the advantage to depend on MPI, while Kokkos does some magic with environment variables (https://github.com/kokkos/kokkos/blob/d54619970c051fcb82ed0a9eab29975998baf832/core/src/impl/Kokkos_CPUDiscovery.cpp#L42).

src/KokkosComm.hpp
aprokop: @dssgabriel There are 3 scenarios, imho.

1. A user uses `KokkosComm::{initialize, finalize}`. In this case, they are forbidden to use their own initialization/finalization.
2. A user may initialize MPI and finalize MPI on their own. If they do that, they are forbidden from calling `KokkosComm::{initialize, finalize}`, and must call `Kokkos::{initialize,finalize}`.
3. A user initialzes and finalizes MPI and Kokkos on their own, and just calls `KokkosComm` routines when necessary.

I disagree with requiring users to call `KokkosComm::{initialize,finalize}` given that it may be a subset of the work, or even optional. I do think that this PR supports all use cases in the current form, as long as we abort if either Kokkos or MPI are initialized in `KokkosComm::initialize`.

dssgabriel: That's clearer, thank you :slightly_smiling_face:

> I disagree with requiring users to call `KokkosComm::{initialize,finalize}` given that it may be a subset of the work, or even optional. I do think that this PR supports all use cases in the current form, as long as we abort if either Kokkos or MPI are initialized in `KokkosComm::initialize`.

I am updating the code accordingly and making the documentation clear about it.

src/KokkosComm.hpp
dssgabriel: Maybe we can strip only `--kokkos-help` for now then.

I agree that if we consider `--help` as a flag from the application, it is up to users to put the necessary guards to prevent all ranks from printing.

dssgabriel:
Last thing we need to think about is: do we abort if we failed to get the desired thread support level?

src/KokkosComm.hpp
```diff
@@ -28,8 +29,49 @@
...
+    if (help_it != argv + argc) {
+      std::swap(*help_it, *(argv + argc - 1));
+      --argc;
+    }
+  }
+  Kokkos::initialize(argc, argv);
+}
+
+inline void initialize(int &argc, char *argv[]) { initialize(argc, argv, MPI_THREAD_MULTIPLE); }
```
devreal: Provide an enum for thread-level that is independent of MPI?

src/KokkosComm.hpp
```diff
@@ -28,8 +29,49 @@
...
+#include <string_view>
+
 namespace KokkosComm {
 
+inline void initialize(int &argc, char *argv[], int mpi_required_thread_lvl) {
+  int flag;
+  MPI_Initialized(&flag);
+  // Eagerly abort if MPI has already been initialized
+  if (0 != flag) {
+    MPI_Abort(MPI_COMM_WORLD, -1);
```
devreal, line 50: Why not allow MPI to be initialized already as long as the requested thread-level matches?

src/KokkosComm.hpp
dssgabriel: Great idea! 👍

src/KokkosComm.hpp
```diff
@@ -28,8 +29,49 @@
 
 #include <Kokkos_Core.hpp>
 
+#include <algorithm>
+#include <cstdio>
+#include <string_view>
+
 namespace KokkosComm {
 
+inline void initialize(int &argc, char *argv[], int mpi_required_thread_lvl) {
```
aprokop:
```suggestion
inline void initialize(int &argc, char *argv[], int mpi_required_thread_lvl = MPI_THREAD_SINGLE) {
```
Providing the default argument would make it equivalent to `MPI_Init` when called without last argument.

src/KokkosComm.hpp
aprokop: Only need to strip `--kokkos-help`. The `--help` flag in the original snippet was part of the application, and it's user responsibility to deal with it. Definitely can't strip any other Kokkos flags.

src/KokkosComm.hpp
aprokop: What purpose would this enum serve? It would require user to not only know MPI names, but also KokkosComm names for different initialization options. The only upside I see if you are looking forward to non-MPI based backends, but I think that's out of scope for this PR.

src/KokkosComm.hpp
dssgabriel: Adding this with the `MPI_THREAD_*` wrapper suggested by Joseph, thank you :+1:

src/KokkosComm.hpp
```diff
@@ -28,8 +29,71 @@
...
+  MPI_Initialized(&flag);
+  // Forbid calling this function if MPI has already been initialized
+  if (0 != flag) {
+    MPI_Abort(MPI_COMM_WORLD, -1);
+  }
+
+  // Forbid calling this function if Kokkos has already been initialized
+  if (Kokkos::is_initialized()) {
+    std::abort();
+  }
```
aprokop, line 56: You don't need this, `Kokkos::initialize()` will do that for you.

src/KokkosComm.hpp
```diff
@@ -28,8 +29,71 @@
...
+  Kokkos::initialize(argc, argv);
+}
+
+inline void initialize(int &argc, char *argv[]) { initialize(argc, argv, ThreadSupportLevel::Multiple); }
+
+inline void finalize() {
+  // Forbid calling this function if Kokkos has already been finalized or isn't yet initialized
+  if (Kokkos::is_finalized() || !Kokkos::is_initialized()) {
+    MPI_Abort(MPI_COMM_WORLD, -1);
+  }
```
aprokop, line 83: Don't need it, `Kokkos::finalize()` will do that for you.

src/KokkosComm.hpp
dssgabriel: I am not sure I understand. _We_ want to abort if `Kokkos::initialize` has already been called. Why try to initialize MPI and Kokkos if we can eagerly prove we got illegally called?

src/KokkosComm.hpp
dssgabriel: Same but for finalization. Also, this lets us cleanly abort on all ranks.

src/KokkosComm.hpp
aprokop: Because Kokkos will do that for you, I believe. I'm not sure MPI will.

devreal: I find it weird that a subproject subsumes the initialization of the main project. I should be able to initialize main Kokkos and KokkosComm from my application. The KokkosComm initializer should initialize main Kokkos if it's not already initialized and otherwise handle that gracefully. Same with MPI above.

aprokop: I don't understand your words about "handle that gracefully". Kokkos initialization will abort if it has been initialized/finalized previously:
https://github.com/kokkos/kokkos/blob/892e13c8c49fa9d4ef9f8dcb69f90d525a6baa58/core/src/impl/Kokkos_Core.cpp#L1056-L1061.
The code inside `KokkosComm::initialize()` checking Kokkos initialization/finalization does not add anything, unless you decide to skip initializing Kokkos, which I argue is a wrong thing to do.

devreal: Well, it adds a totally unexpected abort if Kokkos is already initialized. What is the justification for that? Why do we insist that KokkosComm owns the initialization of Kokkos when we could handle it more gracefully and just not initialize/finalize Kokkos if the application decides to do it on their own before calling KokkosComm::initialize? If Kokkos Remote Spaces started doing the same the two projects could not be used together.

Why would code like this be erroneous?

```C
Kokkos::initialize();
KokkosComm::initialize();

KokkosComm::finalize()?
Kokkos::finalize()?
```

The constraints outlined in https://github.com/kokkos/kokkos-comm/pull/88/files#diff-9933f12ec04a5495a85241b7c9e37827dc1e58f78e034c953a4c83fae4b8416fR47 are unnecessarily strict and the second option actually prevents KokkosComm from initializing any of its own infrastructure. The `is_initialized` routines are there to handle prior init gracefully, instead of just bailing out.

aprokop: > Why would code like this be erroneous?

That is the crux of the matter: this code would be erroneous regardless. You can't call `Kokkos::finalize()` twice, which it will: once inside `KokkosComm::finalize()` and once after. The point I'm trying to make is that if we do optional initialization inside `initialize`, it is not clear who is responsible for finalizing the resource. Unless you carry some state that tells you whether it was `KokkosComm` that initialized the resource and has to finalize it, I don't see it as a good idea.

devreal: > Unless you carry some state that tells you whether it was KokkosComm that initialized the resource and has to finalize it, I don't see it as a good idea.

That is exactly how it's done. It's trivial. I don't know how saving two extra byte of state could possibly justify making the API so much more cumbersome. 10 extra lines of code, two boolean to check whether KokkosComm owns the Kokkos initialization and the MPI initialization. That's it, and the whole discussion in the dox about initialization this way or that way being **forbidden** can go away.

aprokop: This PR does not have the state, only free functions. Introducing the state has its own challenges, I think.

Let's just see what others say. Particularly, would be nice to have feedback from Kokkos developers.

dssgabriel: Enforcing these constraints if users choose our API and handling the initialization process for both MPI and Kokkos lets us guarantee that they are called in the correct order.

The following code is ill-formed, as `Kokkos::initialize` does some special checks for MPI, which is initialized later in `KokkosComm::initialize`:
```cpp
Kokkos::initialize(argc, argv);
KokkosComm::initialize(argc, argv);

KokkosComm::finalize();
Kokkos::finalize();
```
There are a multitude of other scenarios where this could go wrong.

With this PR, users can do **only one** of these two options:
1. Use `KokkosComm::{initialize,finalize}` and let it do all the work for initializing/finalizing MPI and Kokkos in the correct order. If users choose our unified interface, they are forbidden to call MPI's or Kokkos's initialization/finalization functions. Our current implementation aborts if they do, albeit with redundant checks.
2. Use `MPI_{Init,Init_thread,Finalize}` and `Kokkos::{initialize,finalize}` "manually". In this case, users are forbidden to call our `KokkosComm::{initialize,finalize}` interface (similarly, the application will abort if they do), and it is their responsibility to call MPI and Kokkos in the correct order.

We do not prevent users from doing things manually. However, the herein proposed API provides a straightforward and foolproof way of initializing/finalizing their KokkosComm application. If we handle redundant calls "gracefully" regardless, this PR does not add any value and is pretty much useless. Moreover, I don’t understand why we would let people write fundamentally wrong code and silence it. If users are not supposed to initialize/finalize more than once, we must prevent them from doing so. We can also abort with an error message and a meaningful error code.

Finally, I'd like to point out that this is only a first step towards KokkosComm initialization/finalization. Once #68 gets merged, we will have an isolated environment (thanks to MPI Sessions) that lets us lift most of — if not all — the constraints required for now.

aprokop: > `Kokkos::initialize` does some special checks for MPI

Are you sure about this? I only see detection of the local MPI rank based on environment variables, but not calls to any MPI routines.

dssgabriel: Kokkos docs mention one should call `MPI_Init` _before_ `Kokkos::initialize`: https://kokkos.org/kokkos-core-wiki/ProgrammingGuide/Initialization.html?highlight=initialize#initialization

devreal: > We do not prevent users from doing things manually.

You are telling people not to initialize your project if they initialize its dependencies manually. That is flawed.

> However, the herein proposed API provides a straightforward and foolproof way of initializing/finalizing their KokkosComm application.

I am not arguing to take that away. If users want to shift all initialization to KokkosComm they can do so. 

> Moreover, I don’t understand why we would let people write fundamentally wrong code and silence it.

Why is that fundamentally wrong? You can still bail if they did it wrong but there is no reason to always bail even if the configuration they chose is correct (e.g., MPI was initialized with the same thread-level.)

> Finally, I'd like to point out that this is only a first step towards KokkosComm initialization/finalization. Once https://github.com/kokkos/kokkos-comm/pull/68 gets merged, we will have an isolated environment (thanks to MPI Sessions) that lets us lift most of — if not all — the constraints required for now.

Then let's not tell them one thing now and another then. We can get close to the semantics we will have then and users will not have to change their code. Keep in mind that what you are proposing now forces users to change their code such that later when KokkosComm uses Sessions the WPM will be broken and no communication outside of KokkosComm will work anymore. Or KokkosComm will be broken because users stuck to manually initializing MPI to get the WPM. We can do better than that. 

> This PR does not have the state, only free functions. Introducing the state has its own challenges, I think.

The approach I would use is an inline accessor to avoid explicit global variables that need to go into a compilation unit. In some header file:

```C
inline bool& get_state_ref() {
  static bool state = false;
  return state;
}
```

dssgabriel: This PR aimed at unifying two calls — `MPI_{Init,Init_thread}` and `Kokkos::initialize` — into a single one.

> You are telling people not to initialize your project if they initialize its dependencies manually. That is flawed.

It means that instead of having two calls, users may now need a third one.

> Why is that fundamentally wrong? You can still bail if they did it wrong but there is no reason to always bail even if the configuration they chose is correct (e.g., MPI was initialized with the same thread-level.)

OK, that's fair.
Once MPI has been initialized, is it possible to retrieve the current thread support level? I can't find anything in OpenMPI's docs, but this is needed to check that the users' requirements are met in `KokkosComm::initialize` and error out if needed, as you've suggested.

> Keep in mind that what you are proposing now forces users to change their code such that later when KokkosComm uses Sessions the WPM will be broken and no communication outside of KokkosComm will work anymore.

Once #68 gets merged, existing code will most likely have to change anyway. E.g. all of our tests use `MPI_COMM_WORLD`, which won't be valid in the SPM. Besides, we will want to provide some object (`KokkosComm::Context` or `Handle`) that wraps the `MPI_Comm`, `Kokkos::ExecutionSpace`, and maybe the associated stream if comms happen on a device. Our API will then use this object instead of a raw `MPI_Comm`. 
All of our current interfaces are still subject to change. IMO it's ok if we tweak the semantics for now, although I agree it's best to get this PR closer to what we're aiming for in the future. :+1: 

> > This PR does not have the state, only free functions. Introducing the state has its own challenges, I think.
>
> The approach I would use is an inline accessor to avoid explicit global variables that need to go into a compilation unit.

I don't think we need to carry any state for the moment, we're only calling `MPI_Init_thread` and `Kokkos::initialize`. Adding `KokkosComm::is_{initiliazed,finalized}` should be enough for now I think.

devreal: > Once MPI has been initialized, is it possible to retrieve the current thread support level? 

The function you're looking for is `MPI_QUERY_THREAD` :) 

> It means that instead of having two calls, users may now need a third one.

I don't see the problem, esp if parts of my application rely on the WPM anyway and I have to ensure that the WPM is initialized. I cannot rely on KokkosComm to do that if I know that at some point the behavior will change. 

Anyway, maybe this PR isn't ready yet and needs to be tied to #68. Let's discuss this during the Monday call, I will try to join (the MPI Forum potentially conflicts with the call).

cedricchevalier19:
It was decided not to continue and to provide stubs for users to initialize correctly.

PR #90, opened 2024-06-18
resolve bug with KokkosComm::irecv, add to unit testing

nicoleavans:
Previously, irecv was implemented in the KokkosComm namespace erroneously. This bug was not captured via unit testing because unit testing directly called Impl::irecv.


cwpearson (dismissed):

janciesko (dismissed):

janciesko:
`clang-format check`  seems to be failing

unit_tests/test_isendirecv.cpp
```diff
@@ -54,7 +54,7 @@ void test_1d(const View1D &a) {
   } else if (1 == rank) {
     int src = 0;
     MPI_Request req;
-    KokkosComm::Impl::irecv(a, src, 0, MPI_COMM_WORLD, req);
+    KokkosComm::irecv(a, src, 0, MPI_COMM_WORLD, req);
```
devreal, line 57: There is still a difference between isend and irecv: isend returns a request, irecv has an output parameter. Is that on purpose?

nicoleavans: @cwpearson I just saw this, what is your opinion?

cwpearson: We should probably have Irecv produce an output as well. I will accept this for now and open an issue against myself to implement it.

cwpearson: https://github.com/kokkos/kokkos-comm/issues/92

cwpearson (approved):

PR #93, opened 2024-06-18
Replace MPI_Init to MPI_Init_thread when needed

hmt23:


cwpearson:
I think we're currently okay, but are we going to run into a situation where folks who just call MPI_Init and then use us are going to have problems?

dssgabriel:
I think this should go in the documentation regarding initialization. We can change our unit tests to match what is the recommended MPI initialization (either `MPI_Init_thread` w/ `MPI_THREAD_MULTIPLE`, or MPI Sessions).

I am opening an issue for docs on how to initialize KokkosComm applications.

dssgabriel (approved):

PR #96, opened 2024-06-18
docs: how to run clang-format in podman

cwpearson:


aprokop (commented):
Feels like there should be a script file to format all files. This would avoid having to hardcode the paths to format.

dssgabriel (approved):

Issue #104, opened 2024-07-04
OSU latency tests are not strictly equivalent


dssgabriel:
Our raw MPI implementation has a barrier which the KokkosComm one does not. This leads to very different results (at least on my machine).

Compiling the current OSU latency bench in "Release" with MPICH 4.1.1 gives me the following results:
<details>
<summary><b>Current implementation</b></summary>

```
1: Test command: /usr/lib64/mpich/bin/mpiexec "-n" "2" "./perf_test-main"
1: Working Directory: /home/dossantosg/dev/kokkos-comm/build/release/perf_tests
1: Test timeout computed to be: 10000000
1: 2024-07-04T10:05:04+02:00
1: Running ./perf_test-main
1: Run on (16 X 4500 MHz CPU s)
1: CPU Caches:
1:   L1 Data 48 KiB (x8)
1:   L1 Instruction 32 KiB (x8)
1:   L2 Unified 1280 KiB (x8)
1:   L3 Unified 18432 KiB (x1)
1: Load Average: 0.48, 0.55, 0.67
1: ***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
1: -----------------------------------------------------------------------------------------------------------------------
1: Benchmark                                                             Time             CPU   Iterations UserCounters...
1: -----------------------------------------------------------------------------------------------------------------------
1: benchmark_osu_latency_KokkosComm_isendirecv/1/manual_time         0.295 us        0.567 us      2291565 bytes=2
1: benchmark_osu_latency_KokkosComm_isendirecv/2/manual_time         0.300 us        0.579 us      2318277 bytes=4
1: benchmark_osu_latency_KokkosComm_isendirecv/4/manual_time         0.303 us        0.575 us      2344989 bytes=8
1: benchmark_osu_latency_KokkosComm_isendirecv/8/manual_time         0.294 us        0.553 us      2312104 bytes=16
1: benchmark_osu_latency_KokkosComm_isendirecv/16/manual_time        0.281 us        0.543 us      2497346 bytes=32
1: benchmark_osu_latency_KokkosComm_isendirecv/32/manual_time        0.329 us        0.592 us      2187558 bytes=64
1: benchmark_osu_latency_KokkosComm_isendirecv/64/manual_time        0.328 us        0.596 us      2125078 bytes=128
[...]

PR #106, opened 2024-07-08
Increase OSU Latency testing range

nicoleavans:
In reference to issue https://github.com/kokkos/kokkos-comm/issues/104. 

Extend testing range to 256MB.


cwpearson (approved):

cwpearson:
This seems fine to me for now, we can revisit if the automatic perf tests end up taking too long.

PR #110, opened 2024-07-12
Don't flatten include hierarchy on install

cwpearson:
Now build and install match and things are less magical.

### DIFF
--- a/perf_tests/test_main.cpp
+++ b/perf_tests/test_main.cpp
@@ -14,7 +14,7 @@
 //
 //@HEADER
 
-#include "KokkosComm_include_mpi.hpp"
+#include "impl/KokkosComm_include_mpi.hpp"
 
 #include <Kokkos_Core.hpp>
 #include <benchmark/benchmark.h>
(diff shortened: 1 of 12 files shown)


perf_tests/test_main.cpp
```diff
@@ -14,7 +14,7 @@
 //
 //@HEADER
 
-#include "KokkosComm_include_mpi.hpp"
+#include "impl/KokkosComm_include_mpi.hpp"
```
cedricchevalier19, line 17: Perhaps comment on why this `impl/` include is used?

Do we still need this if we do not use MPI compiler wrappers? When using a regular compiler, MPI headers should be found using `-isystem`, so warnings should not be emitted.

cedricchevalier19 (dismissed):

perf_tests/test_main.cpp
cwpearson: Let me revisit this one in https://github.com/kokkos/kokkos-comm/issues/114

cedricchevalier19 (approved):

cwpearson:
I neglected to link the issue I opened before.

PR #118, opened 2024-07-25
Add link to SNL site-specific docs

cwpearson:


cedricchevalier19 (commented):
Ok for me but I would prefer that a sandian gives the green light.

cwpearson:
One of the purposes of gitlab-ex.sandia.gov is to host repositories that non-Sandians can access, so the URL is public information and it's fine to post the link here.

However, we do not expect to give non-Sandians access to that repo, because it contains some internal documentation about how the CI is set up at Sandia, how to compile/run on specific systems on Sandia, etc.

cedricchevalier19 (approved):

PR #125, opened 2024-10-07
Stream-ordered wait{_any,_all}, fix wait_any implementation

cwpearson:
This adds `wait`, `wait_any`, `wait_all` functions like

```c++
template<KokkosComm::KokkosExecutionSpace Space>
KokkosComm::mpi::wait(const Space &exec, KokkosComm:Req<Mpi> req);
```

Semantically, this inserts the "wait" into the provided execution space instance - i.e., existing work in the instance is allowed to complete, and future work must wait for the communication to complete.
The 1-argument version without an execution space just calls the two-argument version with `Kokkos::DefaultExecutionSpace`.

- [ ] documentation
- [ ] more tests

### DIFF
--- a/src/KokkosComm/mpi/req.hpp
+++ b/src/KokkosComm/mpi/req.hpp
@@ -64,12 +64,22 @@ class Req<Mpi> {
  private:
   std::shared_ptr<Record> record_;
 
+  template <KokkosExecutionSpace ExecSpace>
+  friend void wait(const ExecSpace &space, Req<Mpi> req);
   friend void wait(Req<Mpi> req);
+  template <KokkosExecutionSpace ExecSpace>
+  friend void wait_all(const ExecSpace &space, std::vector<Req<Mpi>> &reqs);
   friend void wait_all(std::vector<Req<Mpi>> &reqs);
+  template <KokkosExecutionSpace ExecSpace>
+  friend int wait_any(const ExecSpace &space, std::vector<Req<Mpi>> &reqs);
   friend int wait_any(std::vector<Req<Mpi>> &reqs);
 };
 
-inline void wait(Req<Mpi> req) {
+template <KokkosExecutionSpace ExecSpace>
+void wait(const ExecSpace &space, Req<Mpi> req) {
+  /* Semantically this only guarantees that `space` is waiting for request to complete. For the MPI host API, we have no
+   * choice but to fence the space before waiting on the host.*/
+  space.fence();
   MPI_Wait(&req.mpi_request(), MPI_STATUS_IGNORE);
   for (auto &f : req.record_->postWaits_) {
     f();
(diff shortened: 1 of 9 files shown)


src/KokkosComm/mpi/req.hpp
```diff
@@ -64,34 +64,68 @@ class Req<Mpi> {
...
   }
   req.record_->postWaits_.clear();
 }
 
-inline void wait_all(std::vector<Req<Mpi>> &reqs) {
+inline void wait(Req<Mpi> req) { wait(Kokkos::DefaultExecutionSpace(), req); }
+
+template <KokkosExecutionSpace ExecSpace>
+void wait_all(const ExecSpace &space, std::vector<Req<Mpi>> &reqs) {
+  space.fence();
```
devreal: I wonder whether this fence is necessary. If the semantics of `KokkosExecutionSpace` is such that enqueued operations are ordered then it is sufficient to block the enqueuing thread until the communication is complete. Subsequent submissions are necessarily ordered with the (then-completed) MPI operations.

janciesko: I think we do not need that fence there. There are two more occurrences of superfluous fences.

janciesko: Consider moving the wait_* functions to a separate file.

unit_tests/mpi/test_waitany.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// TODO: test call on no requests
+
+TYPED_TEST(MpiWaitAny, default_execution_space) {
+  wait_any<Kokkos::DefaultExecutionSpace, typename TestFixture::Scalar>();
+}
+
+TYPED_TEST(MpiWaitAny, default_host_execution_space) {
+  if constexpr (std::is_same_v<Kokkos::DefaultHostExecutionSpace, Kokkos::DefaultExecutionSpace>) {
```
janciesko, line 98: Consider adding a MpiWaitAll test.

src/KokkosComm/mpi/req.hpp
cwpearson: Semantically we're trying to order this "wait" after anything else already in the execution space instance. There's no way to actually enqueue an arbitrary host function in a Kokkos execution space instance, so the only way to ensure everything already in the instance is done is to fence it. Then we can call the MPI functions to complete the communication.

unit_tests/mpi/test_waitany.cpp
cwpearson: done

src/KokkosComm/mpi/req.hpp
devreal: The difference is not observable. The operations have been started so they have seen the execution space at the time the operation was started. Subsequent kernels should not interfere with these operations, nor do operations submitted after the `wait` (because they will themselves synchronize the execution space before starting the operation). And I don't think `wait_all` should *guarantee* synchronization of the space because that is more than strictly needed.

dssgabriel:
Some of this was addressed in #216. Currently, the semantics of `Request`s are to synchronize only the communication (which fences on the execution space associated with the comm operation iff there was some unpacking to do in the callbacks).
Can you confirm we can close this @cwpearson?

PR #126, opened 2024-10-07
allow ctest to run from the top level build

cwpearson:


dssgabriel (approved):
Didn't realize my changes would break this! Feel free to remove the comments if this should be obvious.

Also, does this change allow us to remove the `enable_testing()` calls that are in the CMake files of the test directories?

janciesko:
This does not work for me. Should this work?
```
$pwd
/ascldap/users/jciesko/RAndD/kokkos-comm
$ctest -j 3
Test project /ascldap/users/jciesko/RAndD/kokkos-comm
No tests were found!!!
```

How do you all feel about `set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)`?

cwpearson:
The following works for me and seems to work in our CI

```bash
cmake -S ... -B ... -DKokkosComm_ENABLE_TESTS=ON -DKokkos_ROOT=...
cmake --build ...
ctest --test-dir ...
```

PR #127, opened 2024-11-04
refactor!: harmonize syntax & include order of headers

### LINKED ISSUE #114: More graceful way of handling `KokkosComm_include_mpi.hpp`
Did this to silence some warnings, but realy we should be able to avoid this by using system includes properly, etc.

Raise in https://github.com/kokkos/kokkos-comm/pull/110

### LINKED ISSUE #119: Consider renaming `mpi/mpi.hpp`
https://github.com/kokkos/kokkos-comm/blob/a968a13851091aea6d65fd299f0777a307bd8b41/src/KokkosComm/mpi/mpi.hpp

Is this name asking for trouble?

### LINKED ISSUE #120: Fix relative include in mpi.hpp
https://github.com/kokkos/kokkos-comm/blob/a968a13851091aea6d65fd299f0777a307bd8b41/src/KokkosComm/mpi/mpi.hpp#L21

dssgabriel:
Fixes #120.

Removes all instances of relative paths in favor of system include-style syntax: `<KokkosComm/...>`

I think we can also get rid of the `KokkosComm/mpi/impl/include_mpi.hpp` header?


dssgabriel:
Follow-up questions (may be left for another PR):
- do we want to follow standard C++ library design guidelines and use `include` instead of `src` as the base directory (since we're a header-only library)?
- similarly, do we want to use `details` subdirs/namespaces instead of `impl` for implementation details?

See the [Beman Project](https://www.bemanproject.org) — an effort to support open-source, community-driven implementations of standard C++ paper proposals — example library structure for more guidance: https://github.com/beman-project/exemplar

I understand that I may be nitpicking here, feel free to dismiss the ideas if you think they're irrelevant to KokkosComm! :slightly_smiling_face:

cedricchevalier19:
It is not clear that we will be a header-only library forever. It might be interesting to have some precompiled parts to hide some internals in our compiling units.

As for `impl`, it is standard practice for kokkos projects, so I think we should keep it.

perf_tests/mpi/test_main.cpp
```diff
@@ -14,10 +14,9 @@
 //
 //@HEADER
 
-#include "KokkosComm/mpi/impl/include_mpi.hpp"
-
 #include <Kokkos_Core.hpp>
 #include <benchmark/benchmark.h>
+#include <mpi.h>
```
cedricchevalier19, line 17: Are we sure that the custom `include_mpi.hpp` is not needed anymore?

(Note that I am in favor of a more standard way to include mpi)

src/KokkosComm/KokkosComm.hpp
```diff
@@ -16,28 +16,28 @@
 
 #pragma once
 
-#include "fwd.hpp"
+#include <KokkosComm/fwd.hpp>
```
cedricchevalier19: This is not necessary, and it can have some advantages to look for a file with a path relative to this file.

src/KokkosComm/fwd.hpp
```diff
@@ -16,12 +16,13 @@
 
 #pragma once
 
-#include <vector>
-
-#include "concepts.hpp"
+#include <KokkosComm/concepts.hpp>
 #include <KokkosComm/config.hpp>
 
+#include <vector>
```
cedricchevalier19: Usually, we put `std` headers close to the top, like second after the associated header file if it is cpp.

src/KokkosComm/mpi/allgather.hpp
```diff
@@ -16,12 +16,12 @@
 
 #pragma once
 
-#include <Kokkos_Core.hpp>
+#include <KokkosComm/traits.hpp>
+#include <KokkosComm/mpi/impl/pack_traits.hpp>
+#include <KokkosComm/mpi/impl/include_mpi.hpp>
+#include <KokkosComm/mpi/impl/types.hpp>
```
cedricchevalier19: We can (should?) keep it relative.

cedricchevalier19 (commented):
A simpler PR with changes focused on the tests and core files where the relative include is upward will be better suited to solve the issue.

perf_tests/mpi/test_main.cpp
dssgabriel: I haven't touched it in the KokkosComm headers, but including MPI directly using the standard header works fine in tests.

src/KokkosComm/KokkosComm.hpp
dssgabriel: I feel like it would be better to stay consistent and use the same include syntax everywhere. When looking at large "standard C++-style" libraries, they always use the system includes.

See e.g. :
- boost/hana: https://github.com/boostorg/hana/blob/master/include/boost/hana/append.hpp
- beman/execution26: https://github.com/beman-project/execution26/blob/main/include/beman/execution26/detail/basic_operation.hpp

What would be the advantages of keeping relative includes?

src/KokkosComm/fwd.hpp
dssgabriel: Similarly, large "standard C++-style" libraries seem to always include `std` headers _after_ project headers (see examples shared previously).

src/KokkosComm/KokkosComm.hpp
cedricchevalier19: We do not rely on INCLUDE_DIRS.

devreal: I agree with @dssgabriel. The moment a client code brings its own `fwd.hpp` KokkosComm will fall apart. Use proper namespacing for include paths.

cedricchevalier19: To me, it is quite the opposite, including a relative path force to use the correct file. Protecting collision with a directory is good but imperfect; if a user can have a `KokkosComm/fwd.h` in a path before ours, we will use it.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else

It is not `required` by the standard that the compiler deals this way with quotes and brackets, but it was an observation. I was bracket-team in the past, and I think using quotes at the correct places can improve correctness and speed.

src/KokkosComm/fwd.hpp
cedricchevalier19: Some others do the opposite: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes.

dssgabriel:
Finally coming around to finish this PR!

We now follow [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else) and [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) for the order of included headers.

Headers are included in the following order, and separated by a blank line between each level:
1. Associated header (as `KokkosComm` is--currently--a header-only library, we are not concerned by this);
2. C system headers;
3. C++ standard library headers;
4. External library headers (priority is given to C-based libraries, e.g. `mpi.h` comes before `Kokkos_Core.hpp`);
5. KokkosComm headers:
   - Base path with `<>` for any header that is "less deeply nested" than the current header (e.g. `<KokkosComm/concepts.hpp>`);
   - Relative path with `""` for headers that are at the same level or "more deeply nested" than the current header.
6. KokkosComm implementation details headers (same syntax rules as the prior level).

src/KokkosComm/KokkosComm.hpp
```diff
@@ -16,29 +16,28 @@
...
 #include "mpi/handle.hpp"
-#include "mpi/irecv.hpp"
+#include "mpi/send.hpp"
 #include "mpi/isend.hpp"
 #include "mpi/recv.hpp"
+#include "mpi/irecv.hpp"
+#include "mpi/barrier.hpp"
 #include "mpi/reduce.hpp"
+#include "mpi/allgather.hpp"
+#include "mpi/alltoall.hpp"
```
cedricchevalier19: Can we have a simpler order between header files?
Like we have semantic "blocks" separated with space (like all the "send")
Or can we use alphabetical order.

src/KokkosComm/mpi/mpi.hpp
cedricchevalier19, line 1: This file is poorly named I think.

unit_tests/mpi/test_isendrecv.cpp
```diff
@@ -14,10 +14,10 @@
 //
 //@HEADER
 
-#include <gtest/gtest.h>
 #include <type_traits>
 
-#include "KokkosComm/KokkosComm.hpp"
+#include <gtest/gtest.h>
+#include <KokkosComm/KokkosComm.hpp>
```
cedricchevalier19:
```suggestion
#include <gtest/gtest.h>

#include <KokkosComm/KokkosComm.hpp>
```

unit_tests/mpi/test_sendrecv.cpp
```diff
@@ -14,10 +14,10 @@
 //
 //@HEADER
 
-#include <gtest/gtest.h>
 #include <type_traits>
 
-#include "KokkosComm/KokkosComm.hpp"
+#include <gtest/gtest.h>
+#include <KokkosComm/KokkosComm.hpp>
```
cedricchevalier19:
```suggestion
#include <gtest/gtest.h>

#include <KokkosComm/KokkosComm.hpp>
```

cedricchevalier19 (dismissed):

src/KokkosComm/KokkosComm.hpp
dssgabriel: I think semantic blocs (P2P, collectives, etc.) is better. We can use alphabetical order within these too. Will update!

src/KokkosComm/mpi/mpi.hpp
dssgabriel: I agree. What about `mpi_space.hpp`?

unit_tests/mpi/test_sendrecv.cpp
dssgabriel: Since these were tests (which are outside the main project), I did the includes as if this was user code that depends on KokkosComm and imports it. Therefore, KokkosComm falls under the "external libraries" and is included in the same group. Do you prefer to have a line break anyway?

nicoleavans (dismissed):
Assuming `mpi.hpp` will be renamed and semantic blocks will be used to separate header files in `KokkosComm.hpp` (as already indicated in comments), I fully support these changes.

unit_tests/mpi/test_sendrecv.cpp
cedricchevalier19: There are in `unit_tests`, so I think it is still part of the project. But I have no strong opinion.

dssgabriel:
An important question while we are at it: should KokkosComm "export" Kokkos Core? Currently, we include `Kokkos_Core.hpp` in our top-level `KokkosComm/KokkosComm.hpp` header, and I don't think we should. We may use it internally but it shouldn't be part of what we directly expose when including KokkosComm, IMHO.

What do you think @cedricchevalier19 @cwpearson @nicoleavans ?

cedricchevalier19:
> An important question while we are at it: should KokkosComm "export" Kokkos Core? Currently, we include `Kokkos_Core.hpp` in our top-level `KokkosComm/KokkosComm.hpp` header, and I don't think we should. We may use it internally but it shouldn't be part of what we directly expose when including KokkosComm, IMHO.

It is a very good point. Ideally we should not export Kokkos Core, but if we do not do this we are not self usable anymore.

I think we can try to get rid of it and see if it is too annoying for our prospective users.

dssgabriel:
Once merged, closes #119 and #114 as well.

cwpearson:
Thanks @dssgabriel! You also raise some good questions. I don't think we can practically avoid exporting `Kokkos_Core.hpp` because we're going to have some concepts as part of our interface that are specialized on Kokkos::View.

cwpearson (approved):

dssgabriel:
@cwpearson 
> we're going to have some concepts as part of our interface that are specialized on Kokkos::View

Isn't it sufficient to export `Kokkos_Core.hpp` within our headers (e.g. `KokkosComm/concepts.hpp`) but not explicitly in our main `KokkosComm/KokkosComm.hpp` entry point?

Since tests are passing without it in our main header, I think we can merge it like this and revert back to including it if that's too annoying for our users.

Issue #86, opened 2024-06-12
Create containers for CI


cedricchevalier19:
To speed up the CI, especially for GPU-enabled compilations, we can use an ad-hoc image.

Ideally, we need:

- GPU and non-GPU aware MPI (but it should be for tests once we have GPU runners)
- Different flavors of MPI implementations
- Different levels of MPI support: with and without MPI_Sessions, for example (#68)

PR #128, opened 2024-11-14
feat: add NCCL communication space backend

dssgabriel:
This is still very much WIP and cannot build at the moment. I am simply opening the PR so that folks meeting at SC24 can exchange on it. :+1:


src/KokkosComm/fwd.hpp
```diff
@@ -16,16 +16,21 @@
...
+#include <KokkosComm/reduction_op.hpp>
 
 namespace KokkosComm {
+
 #if defined(KOKKOSCOMM_ENABLE_MPI)
 class Mpi;
 using DefaultCommunicationSpace  = Mpi;
 using FallbackCommunicationSpace = Mpi;
+#elif defined(KOKKOSCOMM_ENABLE_NCCL)
+class Mpi;
```
cedricchevalier19, line 21: I think I should depend on MPI in this case.

src/KokkosComm/nccl/req.hpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+inline auto wait_any(std::span<Req<Nccl>> reqs) -> int {
+  for (size_t i = 0; i < reqs.size(); ++i) {
+    auto completed = cudaStreamQuery(&reqs[i].get_inner());
+    if (cudaSuccess == completed) {
+      return true;
+    }
+  }
+  return false;
```
cedricchevalier19: Should you block one request if none of the input is ready?

src/KokkosComm/collective.hpp
```diff
@@ -16,19 +16,35 @@
...
   Impl::Barrier<ExecSpace, CommSpace>{std::forward<Handle<ExecSpace, CommSpace>>(h)};
 }
 
+namespace Experimental {
+
+template <KokkosView SendView, KokkosView RecvView, ReductionOperator RedOp,
+          KokkosExecutionSpace ExecSpace = Kokkos::DefaultExecutionSpace,
+          CommunicationSpace CommSpace   = DefaultCommunicationSpace>
+auto reduce(Handle<ExecSpace, CommSpace>& h, const SendView& sv, RecvView& rv, int root) -> Req<CommSpace> {
+  return Impl::Reduce<SendView, RecvView, RedOp, ExecSpace, CommSpace>(h, sv, rv, root);
```
cedricchevalier19: I do not exactly understand how it will work. Don't we need to call it `execute()`?

src/KokkosComm/nccl/reduce.hpp
```diff
@@ -0,0 +1,92 @@
...
+    space.fence();
+    if ((root == rank) && !KokkosComm::is_contiguous(rv)) {
+      auto recv_args = RecvPacker::allocate_packed_for(space, "reduce recv", rv);
+      space.fence();
+      using SendScalar = typename SendView::non_const_value_type;
+      ncclReduce(send_args.view.data(), recv_args.view.data(), send_args.count, send_args.datatype, op, root, comm,
+                 space.cuda_stream());
+      RecvPacker::unpack_into(space, rv, recv_args.view);
+    } else {
+      space.fence();  // is this fence necessary?
```
cedricchevalier19: I would remove this fence.

src/KokkosComm/nccl/reduce.hpp
```diff
@@ -0,0 +1,92 @@
...
+  } else {
+    using SendScalar = typename SendView::value_type;
+    if ((root == rank) && !KokkosComm::is_contiguous(rv)) {
+      auto recv_args = RecvPacker::allocate_packed_for(space, "reduce recv", rv);
+      space.fence();
+      ncclReduce(sv.data(), recv_args.view.data(), sv.span(), datatype_v<SendScalar>, op, root, comm,
+                 space.cuda_stream());
+      RecvPacker::unpack_into(space, rv, recv_args.view);
+    } else {
+      space.fence();  // is this fence necessary?
```
cedricchevalier19: Not needed in my opinion.

src/KokkosComm/nccl/impl/packer.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <KokkosComm/traits.hpp>
+#include <KokkosComm/impl/KokkosComm_contiguous.hpp>
+#include <KokkosComm/nccl/impl/types.hpp>
+
+#include <nccl.h>
+
+namespace KokkosComm::Experimental::nccl::Impl::Packer {
+
+template <KokkosView View>
+struct NcclArgs {
```
cedricchevalier19: I think `PackedViewNccl` or something like that will be a better name. Looking at the following code, I thought `NCClArgs` was an NCCL thing.

cedricchevalier19 (commented):
Some thoughts.

src/KokkosComm/collective.hpp
dssgabriel: You are right, I fixed this in [`9c185af` (#128)](https://github.com/kokkos/kokkos-comm/pull/128/commits/9c185af34dd8b449d414e72bacdf0602bb8c5a37).

cwpearson:
I think it builds now. My workstation has GPU-aware MPI problems, so I need to fix those before continuing.

dssgabriel:
Just rebased this on top of latest `develop`.

src/KokkosComm/fwd.hpp
dssgabriel: Fixed in https://github.com/kokkos/kokkos-comm/pull/128/commits/68a7609dd1dea37e5d7eeeaceceb9e62f2bc0fe0 :+1:

src/KokkosComm/nccl/req.hpp
dssgabriel: At least for now, I am following what we do for MPI.

src/KokkosComm/nccl/impl/packer.hpp
dssgabriel: Fixed in https://github.com/kokkos/kokkos-comm/pull/128/commits/926438ff81b9d4310e8a279d25dceb6def5be6f4 :+1:

dssgabriel:
@cwpearson Looks like https://github.com/kokkos/kokkos-comm/pull/128/commits/f25f24359bd4fe82c20980f59cb3bafe4ce0e390 breaks a lot of stuff.

Should we prefer KokkosComm custom "marker" structs for our reduction operators instead of trying to use Kokkos Core's built-in reducers?
This is what I initially did, but after a discussion with @cedricchevalier19  (albeit a few months ago), I thought we could use the ones provided by Core instead, but that doesn't seem to work. Compilation crashes as these expect some template parameters, which isn't practical for our users and is unrelated to an MPI/NCCL reduction comm, IMHO.

I reverted in https://github.com/kokkos/kokkos-comm/pull/128/commits/887b606603a31996812115fb8fc1e6f312731091. Tell me if you think it's better to go with Kokkos Core's built-in reducers as "markers".

dssgabriel:
Everything builds after the big refactor! :partying_face: 

The only missing things before merging this should be:
- [ ] update the CI to enable building the full NCCL backend
- [ ] add some more NCCL tests (which should be run in CI?)

CMakeLists.txt
cwpearson, line 1: What's with the (seemingly?) spurious removals in this file?

cwpearson (changes requested):
Generally it looks okay, thanks for the hard work!

CMakeLists.txt
dssgabriel: That is weird. I haven't touch at the top-level CMake since opening the PR so I must have messed something up when rebasing. Will fix!

dssgabriel: Reverted in https://github.com/kokkos/kokkos-comm/pull/128/commits/429c3b07f7d5c6a282590462a122e3bda2b5f6be.

dssgabriel:
Do we want to write the tests and enable the CI for NCCL in this PR or should we break this up into smaller ones?

cwpearson:
> Do we want to write the tests and enable the CI for NCCL in this PR or should we break this up into smaller ones?

If we could have at least one test (even a minimal one) that would be great. Would it be easiest to add a build-only test running under GitHub actions? It would also serve as a minimal example of how to use the API.

Then we can add additional real tests running at SNL or CEA.

cedricchevalier19:
I think only a build test is enough for now.

unit_tests/nccl/test_nccl.cpp
```diff
@@ -0,0 +1,123 @@
...
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
```
cedricchevalier19, line 3: Add a comment that this test is a NCCL smoke test, nothing related to Kokkos-Comm.

unit_tests/CMakeLists.txt
```diff
@@ -76,54 +71,80 @@ if(KOKKOSCOMM_ENABLE_MPI)
...
+    COMMAND
+      ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} 2 ./test-main-mpi
+  )
+endif()
+
+if(KOKKOSCOMM_ENABLE_NCCL)
+  # Standalone NCCL smoke tests (do not use KokkosComm)
+  add_executable(test-nccl)
+  target_sources(test-nccl PRIVATE nccl/test_nccl.cpp)
+  # doesn't use KokkosComm, so explicitly link MPI
```
cedricchevalier19:
```suggestion
  # doesn't use KokkosComm, so explicitly link with NCCL
```

cmake/FindNCCL.cmake
```diff
@@ -0,0 +1,99 @@
...
+      )
+    endif()
+    message(STATUS "NCCL version: ${NCCL_VERSION_FROM_HEADER}")
+  else()
+    message(STATUS "NCCL version < 2.3.5-5")
+  endif()
+  set(CMAKE_REQUIRED_INCLUDES ${OLD_CMAKE_REQUIRED_INCLUDES})
+
+  message(STATUS "Found NCCL (include: ${NCCL_INCLUDE_DIRS}, library: ${NCCL_LIBRARIES})")
+  mark_as_advanced(NCCL_ROOT_DIR NCCL_INCLUDE_DIRS NCCL_LIBRARIES)
```
cedricchevalier19: We should define proper CMake targets instead of relying on variable. (Not mandatory in this PR, can be improved later).

cedricchevalier19 (commented):
If it is not too hard, set up a build test on GitHub. Else, if it is a significant amount of work,  I am OK to merge it as-is and add the CI in another PR.

cwpearson:
> @cwpearson Looks like [f25f243](https://github.com/kokkos/kokkos-comm/commit/f25f24359bd4fe82c20980f59cb3bafe4ce0e390) breaks a lot of stuff.
> 
> Should we prefer KokkosComm custom "marker" structs for our reduction operators instead of trying to use Kokkos Core's built-in reducers? This is what I initially did, but after a discussion with @cedricchevalier19 (albeit a few months ago), I thought we could use the ones provided by Core instead, but that doesn't seem to work. Compilation crashes as these expect some template parameters, which isn't practical for our users and is unrelated to an MPI/NCCL reduction comm, IMHO.
> 
> I reverted in [887b606](https://github.com/kokkos/kokkos-comm/commit/887b606603a31996812115fb8fc1e6f312731091). Tell me if you think it's better to go with Kokkos Core's built-in reducers as "markers".

In today's meeting, we agreed it's fine to move forward with Kokkos Comm tag structs rather than trying to use Kokkos Core ones

cmake/FindNCCL.cmake
cwpearson, line 1: I think we're obliged to respect the pytorch license
* put FindNCCL in its own subdirectory (maybe `cmake/thirdparty/pytorch/...`)
* Copy the Pytorch License into that subdirectory as well - the pytorch license allows reusing and redistributing their code.
* Modify our cmake so it can still find the file in this new location

unit_tests/CMakeLists.txt
cwpearson, line 1: Some of these cmake diffs will probably go away with the new cmake formatting?

cwpearson (changes requested):
Generally looks great, thanks for the hard work. It's a bit tough to review such a big PR, but with these things fixed I will be fine merging it and resolving any issues as they arise.

unit_tests/CMakeLists.txt
dssgabriel: Yes, I will rebase on top of `develop` and hopefully these should disappear.

cmake/FindNCCL.cmake
dssgabriel: Resolved in a74c96da530c1e64b5a65ce2828eeefe3bf5f0a6

dssgabriel:
We still need a compile-only CI workflow before merging, though. Maybe we can simply fetch an MPI binary instead of recompiling our own for the KokkosComm+NCCL build (unlike we do for the MPI workflow).
Ideally, I would like to add unit tests for the collectives as well, so that we have full coverage on the build pipeline.

Hopefully, I will have this ready before Tuesday's meeting next week, so we can discuss it with everyone and decide whether it's best to split it into smaller PRs so multiple people can easily review.
E.g., we could split it as:
1. NCCL infra: CMake stuff, compile-options, CI workflows, test utils, etc. (could be further split up if we want to)
2. Structure specializations
3. P2P interfaces + unit tests
4. Collectives interfaces + unit tests

cwpearson:
Maybe it's easiest to define a simple docker container? e.g. maybe Nvidia provides a base image, then we install nccl and MPI, then run the CI inside the container.

dssgabriel:
> Maybe it's easiest to define a simple docker container? e.g. maybe Nvidia provides a base image, then we install nccl and MPI, then run the CI inside the container.

Yes, this seems to be the way to go.
I am trying to build both MPICH and OpenMPI with CUDA support using a base image from NVIDIA, and generate custom container images that we cache (this ties in with #86). Then, we do build-only jobs of Kokkos-master + KokkosComm_MPI/KokkosComm_NCCL with each MPI implementation. This should significantly speed up the build-only CI and let us test both major MPI vendors in one shot.

The only thing I am wondering is, should we actually build MPIs, or can we directly install binaries that are CUDA-aware? If we do build, do we do it as part of the Linux Compile-Only workflow or separate it? All of this should be in its own PR, though.

cwpearson:
if we can get a CUDA-aware binary install that's completely fine.

dssgabriel:
Some updates on this PR:

1. I have reworked the `FindNCCL.cmake` module following [CMake's documentation](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#find-modules), so we now roll our own instead of using PyTorch's one.
   
   Also, we now expose targets so we can idiomatically link against NCCL with:
   ```cmake
   target_link_libraries(KokkosComm INTERFACE NCCL::NCCL)
   ```

2. I have enabled building KokkosComm with the NCCL backend in the Compile-only workflow, but I still need to update this to run concurrently with the build of KComm on the MPI backend. 
   
   Apart from [this NVHPC image](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/nvhpc?version=25.7-devel-cuda12.9-ubuntu24.04), I could not find pre-built binaries of an MPI impl with CUDA support. I see two options:
   - We use the NVHPC image directly (though I think it's much bigger than the CUDA one we currently use), or;
   - We cache the CUDA-aware Open MPI build and reuse it in this workflow (see #86).
  
   What do you think @cwpearson @cedricchevalier19 ?
  
3. Now that the CI attempts to build the NCCL backend, it looks like I will have a few issues to fix, so I'll be working on that (probably only on the evenings and weekend though, since I have higher priority PhD stuff on the table).

Once everything passes the CI checks, I think we should work on splitting this PR up into reviewable chunks.

dssgabriel:
At last, the compile-only workflow finally passes with the NCCL backend! :tada: 

I will add the remaining tests for the NCCL collectives, and this should be ready to merge.

src/KokkosComm/nccl/recv.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  Req<Nccl> req{space.cuda_stream()};
+  if (is_contiguous(rv)) {
+    ncclRecv(data_handle(rv), span(rv), Impl::datatype_v<T>, peer, comm, space.cuda_stream());
+  } else {
+    using Packer = typename Impl::PackTraits<RecvView>::packer_type;
+    auto args    = Packer::pack(space, rv);
+    // TODO: Consider using a private stream pool to avoid synchronizing the underlying stream, which may not
+    // be empty and have in-flight operations we do not want to wait on.
+    space.fence();  // make sure allocation is complete before receiving
```
cedricchevalier19: I do not think this `fence()` is necessary as `pack()` seems to follow the logic of Kokkos to enqueue its work in the space.

src/KokkosComm/nccl/recv.hpp
```diff
@@ -0,0 +1,54 @@
...
+auto recv(const ExecSpace &space, RecvView &rv, int peer, ncclComm_t comm) -> Req<Nccl> {
+  using T = typename RecvView::non_const_value_type;
+  Kokkos::Tools::pushRegion("KokkosComm::Impl::recv");
+
+  Req<Nccl> req{space.cuda_stream()};
+  if (is_contiguous(rv)) {
+    ncclRecv(data_handle(rv), span(rv), Impl::datatype_v<T>, peer, comm, space.cuda_stream());
+  } else {
+    using Packer = typename Impl::PackTraits<RecvView>::packer_type;
+    auto args    = Packer::pack(space, rv);
```
cedricchevalier19: `Packer::allocate_packed_for` looks more appropriate. We do not need to copy the original view in the reception buffer.

src/KokkosComm/nccl/reduce.hpp
```diff
@@ -0,0 +1,78 @@
...
+                "KokkosComm::Experimental::nccl::reduce: only rank-1 Views are supported");
+  Kokkos::Tools::pushRegion("KokkosComm::Experimental::nccl::reduce");
+
+  Req<Nccl> req{space.cuda_stream()};
+  if (is_contiguous(sv)) {
+    if (rank != root and is_contiguous(rv)) {
+      ncclReduce(data_handle(sv), data_handle(rv), span(sv), Impl::datatype_v<ST>, op, root, comm, space.cuda_stream());
+    } else {
+      auto recv_args = RecvPacker::allocate_packed_for(space, "reduce recv", rv);
+      space.fence();
```
cedricchevalier19:
```suggestion
```

I think we do not need to `fence()`.

src/KokkosComm/nccl/reduce.hpp
```diff
@@ -0,0 +1,78 @@
...
+      ncclReduce(data_handle(sv), data_handle(recv_args.view_), span(sv), Impl::datatype_v<ST>, op, root, comm,
+                 space.cuda_stream());
+      RecvPacker::unpack_into(space, rv, recv_args.view_);
+      req.extend_view_lifetime(recv_args.view_);
+    }
+  } else {
+    auto send_args = SendPacker::pack(space, sv);
+    space.fence();
```
cedricchevalier19:
```suggestion
```

src/KokkosComm/nccl/reduce.hpp
```diff
@@ -0,0 +1,78 @@
...
+    if (rank != root and is_contiguous(rv)) {
+      ncclReduce(data_handle(send_args.view_), data_handle(rv), span(send_args.view_), Impl::datatype_v<ST>, op, root,
+                 comm, space.cuda_stream());
+    } else {
+      auto recv_args = RecvPacker::allocate_packed_for(space, "reduce recv", rv);
+      space.fence();
```
cedricchevalier19:
```suggestion
```

src/KokkosComm/nccl/send.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  Req<Nccl> req{space.cuda_stream()};
+  if (is_contiguous(sv)) {
+    ncclSend(data_handle(sv), span(sv), Impl::datatype_v<T>, peer, comm, space.cuda_stream());
+  } else {
+    using Packer = typename Impl::PackTraits<SendView>::packer_type;
+    auto args    = Packer::pack(space, sv);
+    // TODO: Consider using a private stream pool to avoid synchronizing the underlying stream, which may not
+    // be empty and have in-flight operations we do not want to wait on.
+    space.fence();  // make sure packing is complete before sending
```
cedricchevalier19:
```suggestion
```

cwpearson (approved):

PR #131, opened 2024-12-18
ci: explicitly test against kokkos master for PR tests

cwpearson:
This way we test against whatever the latest release is, and we can add some nightlies to test against `develop`

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -13,7 +13,7 @@ on:
     types: [ opened, reopened, synchronize ]
 
 jobs:
-   Kokkos-040200-OpenMPI-OpenMP-Release:
+   Kokkos-master-OpenMPI-OpenMP-Release:
     env:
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
       - run: echo "This job's status is ${{ job.status }}."
       - name: Build Kokkos
         run: |
-          git clone https://github.com/kokkos/kokkos.git "$KOKKOS_SRC"
+          git clone https://github.com/kokkos/kokkos.git --branch master "$KOKKOS_SRC"
```
cedricchevalier19: I do not think it is useful as the next line does an explicit `git checkout 4.2.00`.

cedricchevalier19 (commented):
I do not think it works.

We can also fetch directly the latest tarball, I will give you a one liner for that.

cedricchevalier19 (approved):
Thanks

PR #158, opened 2025-03-26
Add Matrix-Vector Product example - 1D distribution

Adrien-Tab:
- Add AXPY example to demonstrate the use of KokkosComm
- Add KokkosComm_ENABLE_EXAMPLES option (default OFF)
- ~Modify KokkosComm::wait_all to bypass std::vector~

### DIFF
--- a/examples/01_mvp/01_mvp.cpp
+++ b/examples/01_mvp/01_mvp.cpp
@@ -15,25 +15,28 @@
 //@HEADER
 
 // This example demonstrates how to perform a distributed matrix-vector product (A * x = y)
-// using KokkosComm. The matrix A is distributed among the ranks by blocks of contiguous rows.
+// using KokkosComm. The matrix A is distributed among the ranks in blocks of contiguous rows.
 // Each rank owns a part of the vector x and will communicate it with other ranks step by step.
-// At each step a node communicates with two other nodes to send and receive data.
+// At each step, a node communicates with two other nodes, one to receive the next part of x and
+// the other that needs our part of x.
 
 // Two helper objects are used to manage the distributed work:
-// - RankDims: contains the number of rows, the start and end row indices for each rank
-// - RankCommInfo: contains the ranks to send data to and receive data from
+// - RankDims: contains the number of rows, the start and end row indices for each rank.
+// - RankCommInfo: contains the ranks to send data to and receive data from.
 
-// Other algorithms can be used to distribute the work.
+// Other communication patterns can be used to exchange the vector x between ranks.
 
 #include <KokkosComm/KokkosComm.hpp>
+
 #include <iostream>
+#include <string_view>
 
 struct RankDims {
-  unsigned int nb_rows;
-  unsigned int row_start;
-  unsigned int row_end;
+  int nb_rows;
+  int row_start;
+  int row_end;
 
-  RankDims(long N, int rank, int size) {
+  RankDims(int N, int rank, int size) {
     this->nb_rows   = N / size;
     this->row_start = rank * this->nb_rows;
     this->row_end   = (rank + 1) * this->nb_rows;
@@ -46,12 +49,12 @@ struct RankDims {
 };
 
 struct RankCommInfo {
-  Kokkos::View<unsigned int*> recv_ranks;
-  Kokkos::View<unsigned int*> send_ranks;
+  Kokkos::View<int*> recv_ranks;
+  Kokkos::View<int*> send_ranks;
 
   RankCommInfo(int rank, int size) {
-    this->recv_ranks = Kokkos::View<unsigned int*>("recv_ranks", size);
-    this->send_ranks = Kokkos::View<unsigned int*>("send_ranks", size);
+    this->recv_ranks = Kokkos::View<int*>("recv_ranks", size);
+    this->send_ranks = Kokkos::View<int*>("send_ranks", size);
 
     for (int i = 0; i < size; i++) {
       this->recv_ranks(i) = (rank + i) % size;
@@ -60,56 +63,53 @@ struct RankCommInfo {
   }
 };
 
-// Provide a default value for N
-void checkArgs(long& N) {
-  if (N < 1) {
-    N = 1 << 12;
-  }
-}
-
 template <typename View>
-bool verify_result(View& y, long N) {
-  for (unsigned long i = 0; i < y.extent(0); i++) {
-    if (y(i) != N) {
-      return false;
-    }
-  }
-  return true;
+bool verify_result(View& y, int N) {
+  double N_d       = static_cast<double>(N);
+  const double sum = N_d * (N_d + 1) * (2 * N_d + 1) / 6;
+  int errors       = 0;
+  Kokkos::parallel_reduce(
+      "Verify", y.extent(0), KOKKOS_LAMBDA(const int i, int& err) { err += (y(i) != sum); }, errors);
+
+  return (errors == 0);
 }
 
 int main(int argc, char* argv[]) {
-  long N = -1;
+  int N = 1 << 12;
 
-  for (int i = 0; i < argc; i++) {
-    if (strcmp(argv[i], "-N") == 0 && i + 1 < argc) {
-      N = std::atoi(argv[++i]);
-    }
-    if (strcmp(argv[i], "-h") == 0) {
+  if (argc > 1) {
+    std::string_view arg(argv[1]);
+    if (arg == "-h") {
       std::cout << "KokkosComm dense square matrix-vector product example \n"
                 << "  Usage: " << argv[0] << " [-N <size>] default size is 2^12" << std::endl;
       return 0;
+    } else if (arg == "-N" && argc > 2) {
+      N = static_cast<int>(std::stoi(argv[2]));
     }
   }
-  checkArgs(N);
 
-  int provided;
-  MPI_Init_thread(&argc, &argv, MPI_THREAD_MULTIPLE, &provided);
-  if (provided != MPI_THREAD_MULTIPLE) {
-    throw std::runtime_error("MPI_THREAD_MULTIPLE is needed");
+  if (MPI_Init(&argc, &argv) != MPI_SUCCESS) {
+    std::cerr << "MPI_Init failed" << std::endl;
+    return 1;
   }
 
   Kokkos::initialize(argc, argv);
   {
   using ExecSpace   = Kokkos::DefaultExecutionSpace;
-  using CommSpace   = KokkosComm::DefaultCommunicationSpace;
-  using matrix_type = Kokkos::View<double**, Kokkos::LayoutRight, ExecSpace>;
-  using vector_type = Kokkos::View<double*, Kokkos::LayoutRight, ExecSpace>;
-  using kk_pair     = Kokkos::pair<long, long>;
+  using CommSpace   = KokkosComm::Mpi;
+  using matrix_type = Kokkos::View<double**, ExecSpace>;
+  using vector_type = Kokkos::View<double*, ExecSpace>;
+  using kk_pair     = Kokkos::pair<int, int>;
 
   int rank, size;
   KokkosComm::Handle<ExecSpace, CommSpace> handle;
   rank = handle.rank();
   size = handle.size();
+
+  if (rank == 0) {
+    std::cout << "Running with " << size << " ranks\n"
+              << "Matrix size " << N << " x " << N << std::endl;
+  }
 
   // Helper objects for distributed work
   RankDims dim(N, rank, size);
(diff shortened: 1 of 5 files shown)


examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+    if (strcmp(argv[i], "-h") == 0) {
+      std::cout << "KokkosComm dense square matrix-vector product example \n"
+                << "  Usage: " << argv[0] << " [-N <size>] default size is 2^12" << std::endl;
+      return 0;
+    }
+  }
+  checkArgs(N);
+
+  int provided;
+  MPI_Init_thread(&argc, &argv, MPI_THREAD_MULTIPLE, &provided);
```
cedricchevalier19: Is `MPI_THREAD_MULTIPLE` always necessary? Like for a Serial+Cuda Kokkos configuration, is it required?

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+  if (provided != MPI_THREAD_MULTIPLE) {
+    throw std::runtime_error("MPI_THREAD_MULTIPLE is needed");
+  }
+
+  Kokkos::initialize(argc, argv);
+  {
+  using ExecSpace   = Kokkos::DefaultExecutionSpace;
+  using CommSpace   = KokkosComm::DefaultCommunicationSpace;
+  using matrix_type = Kokkos::View<double**, Kokkos::LayoutRight, ExecSpace>;
```
cedricchevalier19: Is `Kokkos::LayoutRight` mandatory?

I thought that we can have either storage for the local data even if the global matrix is row distributed.

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+  using vector_type = Kokkos::View<double*, Kokkos::LayoutRight, ExecSpace>;
```
cedricchevalier19:
```suggestion
  using vector_type = Kokkos::View<double*, ExecSpace>;
```
Layout is not needed for 1D object.

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+  Kokkos::parallel_for("Initialize", dim.nb_rows, KOKKOS_LAMBDA(const int i) {
+    for (int j = 0; j < N; j++) {
+      A(i, j) = 1.0;
+    }
+    x(i) = 1.0;
+    y(i) = 0.0;
+  });
+
+  // Set communication vector for send/recv operations
+  Kokkos::View<double*, Kokkos::LayoutRight, ExecSpace> comm_vector("comm_vector", dim.nb_rows + size);
```
cedricchevalier19: Why not a `vector_type`?

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+  vector_type x("x", dim.nb_rows);
+  vector_type y("y", dim.nb_rows);
+
+  // Initialize A, x, y
+  Kokkos::parallel_for("Initialize", dim.nb_rows, KOKKOS_LAMBDA(const int i) {
+    for (int j = 0; j < N; j++) {
+      A(i, j) = 1.0;
+    }
+    x(i) = 1.0;
+    y(i) = 0.0;
```
cedricchevalier19:
```suggestion
```

`y` is already zeroed.

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+  RankCommInfo comm_info(rank, size);
+
+  matrix_type A("A", dim.nb_rows, N);
+  vector_type x("x", dim.nb_rows);
+  vector_type y("y", dim.nb_rows);
+
+  // Initialize A, x, y
+  Kokkos::parallel_for("Initialize", dim.nb_rows, KOKKOS_LAMBDA(const int i) {
+    for (int j = 0; j < N; j++) {
+      A(i, j) = 1.0;
```
cedricchevalier19: Perhaps put other value than the same everywhere.

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+  auto span_compute = Kokkos::subview(compute_vector, kk_pair(0, dim.nb_rows));
+  // Span to store the received data for next step
+  auto span_comm = Kokkos::subview(comm_vector, kk_pair(0, dim.nb_rows + size));
+
+  // Copy x to the computation vector for first step
+  Kokkos::deep_copy(span_compute, x);
+  RankDims current_dim = dim;
+
+  // Communication and computation steps
+  for (int step = 1; step < size; step++) {
```
cedricchevalier19, line 157: What is a step in this algorithm?

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+// This example demonstrates how to perform a distributed matrix-vector product (A * x = y)
+// using KokkosComm. The matrix A is distributed among the ranks by blocks of contiguous rows.
+// Each rank owns a part of the vector x and will communicate it with other ranks step by step.
+// At each step a node communicates with two other nodes to send and receive data.
```
cedricchevalier19: Why not doing collective? And can you precise the data you are talking about?

src/KokkosComm/mpi/req.hpp
```diff
@@ -77,12 +77,24 @@ inline void wait(Req<Mpi> req) {
   req.record_->postWaits_.clear();
 }
 
+inline void wait_all() {}
+
+inline void wait_all(Req<Mpi>& req) {
+  wait(req);
+}
+
```
cedricchevalier19: One goal of Kokkos-Comm is to keep the API as small as possible. Can you argument why this function is necessary?

examples/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+    CXX
+  DESCRIPTION "Examples for the KokkosComm experimental communication interfaces"
+)
+
+# Treat the unit tests as a separate project
+# If not included in the Kokkos Comm build, find Kokkos Comm to do a standalone build
+if(NOT TARGET KokkosComm::KokkosComm)
+  find_package(KokkosComm REQUIRED)
+endif()
+
```
cedricchevalier19, line 21: You also have explicit calls to `Kokkos` and `MPI`. I know that not everybody agrees with me, but I think you should do `find_package(MPI REQUIRED)` and `find_package(Kokkos REQUIRED)` as well.

By the way, we may want to get rid of this MPI dependency for other Kokkos-Comm backends.

cedricchevalier19 (changes requested):
I think that the added `wait_all` is not required, and even so, it should be in a separate PR.

src/KokkosComm/mpi/req.hpp
Adrien-Tab: It was meant to allow writing `wait_all(req1, req2, ...)`.
Yes, it's not necessary. I'll revert the changes.

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+    if (strcmp(argv[i], "-N") == 0 && i + 1 < argc) {
+      N = std::atoi(argv[++i]);
+    }
+    if (strcmp(argv[i], "-h") == 0) {
+      std::cout << "KokkosComm dense square matrix-vector product example \n"
+                << "  Usage: " << argv[0] << " [-N <size>] default size is 2^12" << std::endl;
+      return 0;
+    }
+  }
+  checkArgs(N);
```
dssgabriel: This section of code looks messy and unnecessarily complex to me:
1. Why is `N` a `long` instead of an unsigned integer since in `checkArgs` you check for strict positivity (`N > 1`)? I would use `size_t`.
2. Why use C standard library functions instead of C++ `std::string`/`string_view` comparison operator? String views should be the preferred option since they do not need an allocation.
3. Why is the `for` loop starting at 0 instead of 1? `argv[0]` is always the executable name.
4. Why use a `for` loop at all since you only check for one of two possible arguments: `-N` or `-h`?

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+  int provided;
+  MPI_Init_thread(&argc, &argv, MPI_THREAD_MULTIPLE, &provided);
+  if (provided != MPI_THREAD_MULTIPLE) {
+    throw std::runtime_error("MPI_THREAD_MULTIPLE is needed");
+  }
+
+  Kokkos::initialize(argc, argv);
+  {
+  using ExecSpace   = Kokkos::DefaultExecutionSpace;
+  using CommSpace   = KokkosComm::DefaultCommunicationSpace;
```
dssgabriel: Since this example relies on MPI-specific functions (`KokkosComm::mpi::barrier`), I would use the MPI comm space explicitly:
```suggestion
  using CommSpace   = KokkosComm::Mpi;
```

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+  using matrix_type = Kokkos::View<double**, Kokkos::LayoutRight, ExecSpace>;
+  using vector_type = Kokkos::View<double*, Kokkos::LayoutRight, ExecSpace>;
+  using kk_pair     = Kokkos::pair<long, long>;
```
dssgabriel: Why use `long` instead of a plain `int` here? If you absolutely need a 64-bit wide type, I would prefer to have `int64_t` explicitly.

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+    RankDims next_dim(N, recv_comm_rank, size);
+    span_comm = Kokkos::subview(comm_vector, kk_pair(0, next_dim.nb_rows));
+
+    // Start MPI communication (non-blocking)
+    auto req_send = KokkosComm::send(handle, x, send_comm_rank);
+    auto req_recv = KokkosComm::recv(handle, span_comm, recv_comm_rank);
+
+    // Compute with current data while communication may happen in the background
+    Kokkos::parallel_for("MatrixVectorProduct", dim.nb_rows, KOKKOS_LAMBDA(const int i) {
+      for (unsigned int j = 0; j < current_dim.nb_rows; j++) {
```
dssgabriel: Why use an `unsigned int` in this loop but not in the other ones?
I would suggest `size_t` instead.

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    // Copy the received data to span_compute for the next step
+    span_compute = Kokkos::subview(compute_vector, kk_pair(0, next_dim.nb_rows));
+    Kokkos::deep_copy(span_compute, span_comm);
+    current_dim = next_dim;
+  }
+
+  // Last step
+  Kokkos::parallel_for("MatrixVectorProduct tail", dim.nb_rows, KOKKOS_LAMBDA(const int i) {
+    for (unsigned int j = 0; j < current_dim.nb_rows; j++) {
```
dssgabriel: Same thing here for `j`.

examples/01_mvp/01_mvp.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+// Other algorithms can be used to distribute the work.
+
+#include <KokkosComm/KokkosComm.hpp>
+#include <iostream>
+
+struct RankDims {
+  unsigned int nb_rows;
+  unsigned int row_start;
+  unsigned int row_end;
```
dssgabriel: I suggest `size_t` instead of `unsigned int` here. If you absolutely need 32-bit wide types, use `uint32_t`.

dssgabriel (changes requested):
Just a quick pass, I agree with all of Cedric's remarks.

I have made additional notes regarding the "parsing" of CLI options that I find too convoluted, and the choice of integer types that looks a bit random to me. I would like to have consistent typing, e.g. always use `size_t` for unsigned stuff and `int` everywhere else. If you require specific bit widths, make it clear by using the types provided by the `cstdint` header.

examples/01_mvp/01_mvp.cpp
Adrien-Tab: Yes, we don't need `MPI_THREAD_MULTIPLE`. In this case, `MPI_THREAD_SINGLE` seems appropriate, as the calls are never in parallel regions.

dssgabriel: Could you replace it with a naked `MPI_Init()` instead?

Adrien-Tab: yes of course

examples/01_mvp/01_mvp.cpp
Adrien-Tab: done it

examples/01_mvp/01_mvp.cpp
Adrien-Tab: yes, I changed it

examples/01_mvp/01_mvp.cpp
Adrien-Tab: Here it's a step of calculations and communications. There's a communication for the next chunk of the distributed x vector. Step 1 performs the calculation on the vector x local to the rank.

examples/01_mvp/01_mvp.cpp
Adrien-Tab: The vector x now corresponds to `x = (1,2, ..., N)` and the matrix A is filled as `A(i,j) = j + 1.0`
which means that each element of `y` is equal to the sum of squares from 1 to N. This is now less trivial, but it can be improved.

examples/01_mvp/01_mvp.cpp
Adrien-Tab: In this example, we progress diagonally, starting with the part of x that is assigned to the rank.

examples/01_mvp/01_mvp.cpp
```diff
@@ -60,57 +63,54 @@ struct RankCommInfo {
...
-    }
-    if (strcmp(argv[i], "-h") == 0) {
+  if (argc > 1) {
+    std::string_view arg(argv[1]);
+    if (arg == "-h") {
       std::cout << "KokkosComm dense square matrix-vector product example \n"
                 << "  Usage: " << argv[0] << " [-N <size>] default size is 2^12" << std::endl;
       return 0;
+    } else if (arg == "-N" && argc > 2) {
+      N = static_cast<int>(std::stoi(argv[2]));
```
dssgabriel: I don't think the `static_cast` is necessary here.

examples/01_mvp/01_mvp.cpp
Adrien-Tab: done

cedricchevalier19:
I'll temporarily close this PR waiting to some API changes to fit more adequately to this usage.

PR #161, opened 2025-06-07
Feature: allreduce

dutkalex:
This PR adds implementation for MPI_Allreduce in its scalar and 1D version, both in-place and with separate input and output views.

### DIFF
--- a/src/KokkosComm/CMakeLists.txt
+++ b/src/KokkosComm/CMakeLists.txt
@@ -62,6 +62,7 @@ if(KOKKOSCOMM_ENABLE_MPI)
         mpi/send.hpp
         mpi/allgather.hpp
         mpi/alltoall.hpp
+        mpi/allreduce.hpp
         mpi/reduce.hpp
         mpi/barrier.hpp
   )
--- a/src/KokkosComm/KokkosComm.hpp
+++ b/src/KokkosComm/KokkosComm.hpp
@@ -35,6 +35,7 @@
 #include "mpi/send.hpp"
 
 #include "mpi/allgather.hpp"
+#include "mpi/allreduce.hpp"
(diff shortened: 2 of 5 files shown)


cwpearson (changes requested):
I think we need to adjust this to fix the install:

https://github.com/kokkos/kokkos-comm/blob/ae4a704313a9040c8db24bbc2008ad644b7e6258/src/KokkosComm/CMakeLists.txt#L54

dutkalex:
> I think we need to adjust this to fix the install:

Problem solved, thanks for the insight!

dssgabriel:
I would prefer not to throw exceptions and rather have a direct `std::exit` call with an error message to let the user know that what they are trying to do is not supported (yet).

There have been discussions on error handling (see #29) but nothing as been decided for now and some people are against the use of exceptions in KokkosComm.

dutkalex:
> I would prefer not to throw exceptions and rather have a direct `std::exit` call with an error message to let the user know that what they are trying to do is not supported (yet).

I have personally no strong opinions about this topic and I'd be completely ok with your proposed change. However, exceptions are already sprinkled in lots of places in the `src/KokkosComm/mpi` files and there is a consistency argument in favor of exceptions here. It would be wiser to have a separate PR implement this change across all the files IMO.

> There have been discussions on error handling (see #29) but nothing as been decided for now and some people are against the use of exceptions in KokkosComm.

It's not clear to me reading this thread if some kind of consensus has emerged on the topic. From all the arguments put forward, it seems to me that the only sensible thing to do upon error is indeed to abort.

cwpearson (approved):
I don't have a strong feeling about the error handling since it's not recoverable either way.

PR #162, opened 2025-06-07
Feature: broadcast

dutkalex:
This PR adds basic implementation for MPI_Bcast, for both scalars and 1D contiguous views.

### DIFF
--- a/src/KokkosComm/CMakeLists.txt
+++ b/src/KokkosComm/CMakeLists.txt
@@ -64,6 +64,7 @@ if(KOKKOSCOMM_ENABLE_MPI)
         mpi/alltoall.hpp
         mpi/reduce.hpp
         mpi/barrier.hpp
+        mpi/broadcast.hpp
   )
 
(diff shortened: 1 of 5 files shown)


cwpearson (changes requested):
I think you need to adjust this to fix the install

https://github.com/kokkos/kokkos-comm/blob/ae4a704313a9040c8db24bbc2008ad644b7e6258/src/KokkosComm/CMakeLists.txt#L54

dutkalex:
> I think you need to adjust this to fix the install

Thanks for the insight. Problem solved :)

dssgabriel:
Same as for the `allreduce` PR: I would prefer not to throw exceptions and rather have a direct `std::exit` call with an error message to let the user know that what they are trying to do is not supported (yet).

cwpearson (approved):
I don't have a strong feeling about the error handling since it's not recoverable either way. We can make it more consistent later.

PR #163, opened 2025-06-10
Feature: Enable different error handling strategies

dutkalex:
This PR proposes to extract into a single place all the runtime checks. This change would:
1. Simplify control flow, remove duplicate patterns and improve readability overall
2. Improve flexibility by making it easier to implement alternative error-handling strategies, as it seems there is no obvious "one size fits all" solution to this question (for reference: #29 #161)


src/KokkosComm/mpi/impl/error_handling.hpp
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include <mpi.h>
+
+namespace KokkosComm::mpi {
+inline void fail_if(bool condition, const char* error_msg) {
+  if (condition) {
+#ifdef KOKKOSCOMM_ABORT_ON_ERROR
+    MPI_Abort(MPI_COMM_WORLD, EXIT_FAILURE);
+#else
```
cedricchevalier19: You should print the error message also in this branch.

src/KokkosComm/mpi/allgather.hpp
```diff
@@ -36,12 +37,9 @@ void allgather(const SendView &sv, const RecvView &rv, MPI_Comm comm) {
   static_assert(KokkosComm::rank<SendView>() <= 1, "allgather for SendView::rank > 1 not supported");
   static_assert(KokkosComm::rank<RecvView>() <= 1, "allgather for RecvView::rank > 1 not supported");
 
-  if (!KokkosComm::is_contiguous(sv)) {
-    throw std::runtime_error("low-level allgather requires contiguous send view");
-  }
-  if (!KokkosComm::is_contiguous(rv)) {
-    throw std::runtime_error("low-level allgather requires contiguous recv view");
-  }
+  ::KokkosComm::mpi::fail_if(!KokkosComm::is_contiguous(sv), "low-level allgather requires contiguous send view");
```
cedricchevalier19:
```suggestion
  KokkosComm::mpi::fail_if(!KokkosComm::is_contiguous(sv), "low-level allgather requires contiguous send view");
```
To be coherent with other parts of the code.

src/KokkosComm/mpi/impl/error_handling.hpp
```diff
@@ -0,0 +1,31 @@
...
+    throw std::runtime_error{error_msg};
```
cedricchevalier19: Can't we use `Kokkos::abort()`? This way we will coherent with the other Kokkos based libraries.

src/KokkosComm/mpi/impl/error_handling.hpp
```diff
@@ -0,0 +1,31 @@
...
+//@HEADER
+
+#pragma once
+
+#include <mpi.h>
+
+namespace KokkosComm::mpi {
+inline void fail_if(bool condition, const char* error_msg) {
+  if (condition) {
+#ifdef KOKKOSCOMM_ABORT_ON_ERROR
```
cedricchevalier19, line 25: Where this macro is defined?

cedricchevalier19 (commented):
Quick comments, from my phone so I am not confident in spotting all the things.

I think having a centralized error handling is a great addition.

src/KokkosComm/mpi/impl/error_handling.hpp
dutkalex: In which branch? Do you mean instead of `MPI_Abort` or instead of throwing?

src/KokkosComm/mpi/impl/error_handling.hpp
dutkalex: There should be a corresponding CMake option, but it looks like I forgot to add it

dutkalex: I've just updated the PR to add it

src/KokkosComm/mpi/impl/error_handling.hpp
dssgabriel: This would be instead of throwing. If Kokkos then throws in its `abort` impl, at least it is coherent at the library level.

We may revert this back once we decide how to handle errors, but for now too many people are against exceptions within this project as it is not clear how to correctly handle them in a distributed context.

cwpearson (approved):
I'm good with it for now -- in the future, do you think it would make sense do this as a macro so we can inject file and line numbers into error messages?

dutkalex:
> Do you think it would make sense do this as a macro so we can inject file and line numbers into error messages?

Hmm... I had not though about that but it would indeed be useful. I guess we could convert the `fail_if` function to a function-like macro without too much trouble and have the client code remain basically unchanged, except for the namespace qualification.

Since this is private code, another approach could also be to extend the argument list and pass `__LINE__` and `__FILE__` at the call site

PR #156, opened 2025-03-10
Persistent API

nicoleavans:
This PR introduces a minimal implementation of persistence functionality to obtain feedback, suggestions, and requests from developers and users. A version utilizing [MPI windows](https://www.mpi-forum.org/docs/mpi-2.2/mpi22-report/node231.htm) is in progress, as are some additional unit tests and a performance test. 

```bash
10/11 Test #10: test-channel .....................   Passed    2.91 sec
```

PR #169, opened 2025-07-29
Add HIP + MPI unit test

nicoleavans:
This adds a unit test equivalent to [test_mpi_cuda_sendrecv.cpp](https://github.com/kokkos/kokkos-comm/blob/develop/unit_tests/mpi/test_mpi_cuda_sendrecv.cpp), verifying that GPU-aware MPI is operating as expected if HIP is enabled.


cwpearson (changes requested):
I think you need to modify CMake to actually build and run this

nicoleavans:
@cwpearson I'll fix that, and I think I also need to remove those references in #156. Thanks for catching that!

cedricchevalier19 (approved):
Looks OK but I do not have a HIP device to try it.

cwpearson (approved):

PR #170, opened 2025-09-09
ci: macOS: don't install cmake

cwpearson:
CMake appears to come pre-installed now.


cwpearson:
The macOS CI tests pass so I am going to merge this. Any other problems can be fixed separately.

PR #172, opened 2025-09-16
ci: replace linting workflows with pre-commit checks

dssgabriel:
This PR is based on #171 (we will need to merge that first) and replaces the formatting and spell-check workflows with one that runs all pre-commit hooks defined and the above-mentioned PR.

Hopefully, this will streamline checks in the CI pipeline and also prevent trailing white space, EOFs, etc., from getting mistakenly introduced in the project.


cwpearson:
In principle, I am okay with this. Can you help me understand - does this actually modify code in the PR, or just post suggestions?

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,58 @@
...
+    branches:
+      - develop
+
+jobs:
+  pre-commit-push:
+    name: Pre-commit on Push
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' }}
+    steps:
+      - uses: actions/checkout@v4
```
cwpearson: let's pin everything to a SHA instead of a tag.

cwpearson (changes requested):

.github/workflows/pre-commit.yml
dssgabriel: Sure!

As an aside, I see SHAs being used in place of tags in many places; is there any reason why folks prefer raw commit SHAs rather than tags? I feel like tags should be easier to maintain/keep track of?

cwpearson: Basically if someone takes over the repo that defines the action, they can just change the tag to point at a new SHA that does bad things. Can't do the same with a commit SHA, since changing the commit changes the SHA.

cwpearson: It is more annoying to maintain, but...we shouldn't need to update the actions very often 😄

dssgabriel: Thanks for the answer! Updating everything to SHAs and rebasing on top of `develop` :+1:

dssgabriel:
> does this actually modify code in the PR, or just post suggestions?

The `reviewdog` action will only post the relevant suggestions for the pre-commit checks that did not pass and **will not** modify code in the PR.

If you have the pre-commit hooks enabled on your fork/clone of the repo, this CI workflow should be guaranteed to pass as the checks and automatic fixes would have been applied locally on your machine at commit-time.

dssgabriel:
Was there anything else besides that SHAs that you wanted me to change before we merge this PR @cwpearson ? If not, I think we can go ahead with it :)

cedricchevalier19:
@dssgabriel Can you describe what is the expected results. I have quite a hard time how something in the CI can be "pre-commit".

dssgabriel:
> Can you describe what is the expected results. I have quite a hard time how something in the CI can be "pre-commit".

This new workflow runs the same _checks_ defined in the pre-commit configuration (but not actual pre-commit hooks) and reports any failures. All our lint checks are thus grouped into this one workflow. Contributors who use the pre-commit hooks added in #171 should have this workflow guaranteed to pass. I can rename it if you feel like it is confusing (though that is the name I have seen in other projects that rely on the same mechanism).

cedricchevalier19 (approved):
Looks interesting, thanks @dssgabriel

cedricchevalier19:
> I can rename it if you feel like it is confusing (though that is the name I have seen in other projects that rely on the same mechanism).

Thanks for the explanation. We can leave the name.

cwpearson (approved):
Great!

PR #175, opened 2025-10-02
fix(mpi): ensure `wait_any` completes at least one request

### LINKED ISSUE #173: mpi: `wait_any` may return without completing at least one request
The current implementation of our MPI `wait_any` wrapper may return without actually completing at least one request:
https://github.com/kokkos/kokkos-comm/blob/45aa5e02d8beb47d4a04a8b3b29d2d33cc09e4ed/src/KokkosComm/mpi/req.hpp#L86-L95

Here, we only do a single pass over the requests. If none of them has completed, we return -1 without having actually waited for any request completion. This does not match the semantics of `MPI_Waitany`, which guarantees the completion of one or more requests upon return (see, e.g., [Open MPI's docs](https://docs.open-mpi.org/en/v5.0.x/man-openmpi/man3/MPI_Waitany.3.html)).

Additionally, the current implementation does not actually wait on completed requests and only tests them. If any callbacks were registered for post-wait execution, these are never called, nor cleared.

The algorithm should rather be something along these lines:
```lua
completed_reqs = 0
-- Keep looping over requests while none has completed
while completed_reqs == 0
  for req in requests
    res = test(req)
    if res == COMPLETED_FLAG
      completed_reqs += 1
      wait(req) -- will execute any registered callbacks
    end
  end
end
```

dssgabriel:
This PR closes #173 by changing the implementation of `Req<Mpi>::wait_any` to actually complete at least one request (and calls any registered callbacks on that request) before returning.

Also includes:
- Changed `wait` to take `KokkosComm::Req<Mpi>` by reference to avoid a potentially costly deep copy of the `postWaits_` callback vector
- Added an overload on `wait` taking an rvalue-reference so that writing the following remains valid:
  ```cpp
  KokkosComm::wait(KokkosComm::send(handle, view, peer));
  ```
- Modernized `wait_all` and `wait_any` interfaces to use `std::span` instead of `std::vector&` references

### DIFF
--- a/src/KokkosComm/mpi/req.hpp
+++ b/src/KokkosComm/mpi/req.hpp
@@ -70,15 +70,17 @@ inline void wait_all(std::vector<Req<Mpi>> &reqs) {
   }
 }
 
-inline int wait_any(std::vector<Req<Mpi>> &reqs) {
-  for (size_t i = 0; i < reqs.size(); ++i) {
-    int completed;
-    MPI_Test(&(reqs[i].mpi_request()), &completed, MPI_STATUS_IGNORE);
-    if (completed) {
-      return true;
+inline void wait_any(std::span<Req<Mpi>> reqs) {
+  while (true) {
+    for (Req<Mpi> &req : reqs) {
+      int flag;
+      MPI_Test(&(req.mpi_request()), &flag, MPI_STATUS_IGNORE);
+      if (flag) {
+        wait(req);
+        return;
+      }
     }
   }
-  return false;
 }
 
 }  // namespace KokkosComm
--- a/src/KokkosComm/mpi/req.hpp
+++ b/src/KokkosComm/mpi/req.hpp
@@ -3,8 +3,9 @@
 
 #pragma once
 
-#include <vector>
 #include <functional>
+#include <span>
+#include <vector>
 
 #include <mpi.h>
 
@@ -51,12 +52,12 @@ class Req<Mpi> {
  private:
   std::shared_ptr<Record> record_;
 
-  friend void wait(Req<Mpi> req);
-  friend void wait_all(std::vector<Req<Mpi>> &reqs);
-  friend int wait_any(std::vector<Req<Mpi>> &reqs);
+  friend void wait(Req<Mpi> &req);
+  friend void wait_all(std::span<Req<Mpi>> reqs);
+  friend void wait_any(std::span<Req<Mpi>> reqs);
 };
 
-inline void wait(Req<Mpi> req) {
+inline void wait(Req<Mpi> &req) {
   MPI_Wait(&req.mpi_request(), MPI_STATUS_IGNORE);
   for (auto &f : req.record_->postWaits_) {
     f();
@@ -64,7 +65,7 @@ inline void wait(Req<Mpi> req) {
   req.record_->postWaits_.clear();
 }
 
-inline void wait_all(std::vector<Req<Mpi>> &reqs) {
+inline void wait_all(std::span<Req<Mpi>> reqs) {
   for (Req<Mpi> &req : reqs) {
     wait(req);
   }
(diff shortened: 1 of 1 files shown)


src/KokkosComm/mpi/req.hpp
```diff
@@ -64,34 +65,40 @@ class Req<Mpi> {
...
+  int completed = 0;
+  while (completed != 0) {
+    for (Req<Mpi> &req : reqs) {
+      int flag;
+      MPI_Test(&(req.mpi_request()), &flag, MPI_STATUS_IGNORE);
+      if (flag) {
+        completed++;
+        wait(req);
+      }
     }
```
cedricchevalier19: It is semantically correct now but I am not really fond of the active wait. Can't we build a MPI_REQUEST array and use `MPI_Waitany`?

dssgabriel: Sure! I think we should do the same for `wait_all` in that case :+1:

dssgabriel: How do we go about calling the registered callbacks on the requests that have completed, though?

cwpearson: `MPI_Waitany` provides the index of the thing that completed, so I think you'd just have to go back and do the callbacks of our associated request.

dssgabriel:
I changed the implementation to match the `MPI_Waitany` semantics, i.e., complete one request and exit, as opposed to what I had before, which was more similar to `MPI_Waitsome` (which may complete multiple requests before returning).

I think we can merge this as it is, and I'll do a follow-up PR that updates our `wait_all` and `wait_any` implementation to extract arrays of `MPI_Request`s and directly call the MPI functions.

src/KokkosComm/mpi/req.hpp
```diff
@@ -64,34 +65,39 @@ class Req<Mpi> {
...
-    int completed;
-    MPI_Test(&(reqs[i].mpi_request()), &completed, MPI_STATUS_IGNORE);
-    if (completed) {
-      return true;
+inline void wait_any(std::span<Req<Mpi>> reqs) {
+  while (true) {
+    for (Req<Mpi> &req : reqs) {
+      int flag;
+      MPI_Test(&(req.mpi_request()), &flag, MPI_STATUS_IGNORE);
+      if (flag) {
```
cedricchevalier19, line 85: Does this work for `MPI_REQUEST_NULL`?
If not, we can end-up with an infinite loop.

src/KokkosComm/mpi/req.hpp
```diff
@@ -64,34 +65,39 @@ class Req<Mpi> {
...
 }
 
-inline int wait_any(std::vector<Req<Mpi>> &reqs) {
-  for (size_t i = 0; i < reqs.size(); ++i) {
-    int completed;
-    MPI_Test(&(reqs[i].mpi_request()), &completed, MPI_STATUS_IGNORE);
-    if (completed) {
-      return true;
+inline void wait_any(std::span<Req<Mpi>> reqs) {
+  while (true) {
```
cedricchevalier19, line 81:
```suggestion
  // FIXME: Active wait-loop
  while (true) {
```

src/KokkosComm/mpi/req.hpp
dssgabriel: I will take a look at what the standard says and fix this accordingly.

cedricchevalier19: > A call to MPI_TEST returns flag = true if the operation identified by request is complete. In such a case, the status object is set to contain information on the completed operation; if the communication object was created by a nonblocking send or receive, then it is deallocated and the request handle is set to MPI_REQUEST_NULL. The call returns flag = false, otherwise. In this case, the value of the status object is undefined. MPI_TEST is a local operation.

I am not an MPI expert but my understanding is that `flag = false` if the input is `MPI_REQUEST_NULL` as it cannot be completed.

cedricchevalier19 (approved):
I approve as I think even if this solution is not perfect it is significantly better.

Moreover a better fix should be common to all wait functions and the API might need quite an evolution to be closer to MPI.

PR #176, opened 2025-10-05
chore: update file headers

dssgabriel:
In alignment with preparations for Kokkos 5.0, update our headers to the same minimalist version as the main Kokkos project (see https://github.com/kokkos/kokkos/pull/8496):
```
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
```

Additionally, I:
- Updated our main LICENSE file to remove the old Kokkos v4.0 header and fix the project name (was "Kokkos MPI")
- Removed copyright license headers in all other files that are not KokkosComm source code (same as Kokkos)
- ~~Changed `#pragma once` in C++ headers to PP guards (same style as used in Kokkos)~~
- Tweaked some whitespace to make it consistent across all files

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,19 +1,3 @@
-#@HEADER
-# ************************************************************************
-#
-#                        Kokkos v. 4.0
-#       Copyright (2022) National Technology & Engineering
-#               Solutions of Sandia, LLC (NTESS).
-#
-# Under the terms of Contract DE-NA0003525 with NTESS,
-# the U.S. Government retains certain rights in this software.
-#
-# Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
-# See https://kokkos.org/LICENSE for license information.
-# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-#
-#@HEADER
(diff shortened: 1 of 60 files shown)


cedricchevalier19 (commented):
Why changing the `#pragma once`?

No licence mentionned in cmake files?

dssgabriel:
I tried to match what Kokkos core does. The main project uses PP guards instead of `pragma once`, and does not have license headers in their CMakeLists (they only seem to be included in the source files?).

cedricchevalier19:
`Kokkos-"Core"` has a coding style that has some legacy so I do not think we have to follow all of it.

To me, the pragma, while not standard, are easier to use in a young project where files might be moved around quite a bit. Else, we need to ensure that the name of the guards are correct in the CI.
Please, let's focus only on the Copyright headers in this PR.

dssgabriel:
It appears that the CMake file for unit tests got reformatted. It looks like it was missed by the PRs related to pre-commit checks, though.

cedricchevalier19 (approved):

Issue #108, opened 2024-07-10
Clarify semantics of `ExecSpace` parameter for communication interfaces


dssgabriel:
KokkosComm currently exposes communication functions with an `ExecSpace` parameter whose purpose is not semantically clear, neither in the code nor in the documentation, and that doesn't map to the behavior we expect.

The documentation for, e.g. `KokkosComm::isend`, states:
> ```cpp
> template<CommunicationMode SendMode, KokkosExecutionSpace ExecSpace, KokkosView SendView>
> Req isend(const ExecSpace &space, const SendView &sv, int dest, int tag, MPI_Comm comm)
> ```
>
> Parameters:
> - `space`: The execution space to operate in
> ...
>
> Template Parameters:
> - `ExecSpace`: A Kokkos execution space to operate
> ...

This lets users rightfully think that the `space` parameter is used for specifying in which execution space the communication has to happen, not the one they must sync with for their communication to be correctly processed.
However, the actual implementation of `KokkosComm::isend` (for contiguous views that don't require packing) does something like:
```cpp
  space.fence();  // can't issue isend until work in space is complete
  MPI_Isend(/*...*/);
```

Why can't we submit more work in the space? What does this fence has to do with the execution space in which the communication operates?

Let's demonstrate these unclear semantics with a code example:
```cpp
// Partition the execution space into two
auto instances = Kokkos::partition_space(space, 1, 1);
auto compute_space = instances[0];
auto comm_space = instances[1];

// Dispatch some parallel work that prepares our data `v` on the compute exec space
Kokkos::parallel_for(Kokkos::RangePolicy(compute_space, 0, v.extent(0)), [=](int const i) {
  v(i) = i;
});

// Send the prepared data using the communication exec space
KokkosComm::isend(comm_space, v, dest, tag, comm).wait(); // WARNING: `v` may not have been ready!
```

In this snippet, we split an execution space into two. Because we're specifying the `compute_space` execution space to the `parallel_for`, it is asynchronous and we _must_ fence on it so that `v` is done initializing _before_ calling `KokkosComm::isend`.
However, the implementation of the latter fences on the given execution space (here, `comm_space`) before doing the actual send. In this example, this is the wrong space to fence on, **and** the fence is useless.

Passing an execution space can be useful — e.g. for specifying where to pack a non-contiguous view, or use a particular CUDA stream once we have the NCCL backend — but it shouldn't be just for us to manually fence on an execution space that may have nothing to do with the view we operate on.

[...]

PR #177, opened 2025-10-06
fix(mpi): make `mpi::send` w/ implicit ExecSpace directly call explicit overload 

### LINKED ISSUE #174: mpi: `send` with implicit ExecSpace does not observe the specified comm mode
Our implementation of `KokkosComm::mpi::send` without a `KokkosExecutionSpace` template parameter does not observe the specified `CommunicationMode`:
https://github.com/kokkos/kokkos-comm/blob/45aa5e02d8beb47d4a04a8b3b29d2d33cc09e4ed/src/KokkosComm/mpi/send.hpp#L32-L56

We should use the `mpi_send_fn` variable defined above, which selects the MPI call corresponding to the specified communication mode, rather than calling `MPI_Send` directly.

Alternatively, we could call the version with an explicit `KokkosExecutionSpace` and construct the default ExecSpace in place.

dssgabriel:
This PR closes #174 by modifying the `KokkosComm::mpi::send` overload, which previously did not take an explicit `ExecSpace` template parameter, to directly invoke the fully explicit implementation, passing a `Kokkos::DefaultExecutionSpace` argument.

This effectively removes the bug noted in the above-mentioned issue, where this overload did not observe the Communication Mode requested by the user at the function call level.

### DIFF
--- a/src/KokkosComm/mpi/send.hpp
+++ b/src/KokkosComm/mpi/send.hpp
@@ -51,6 +51,7 @@ void send(const ExecSpace &space, const SendView &sv, int dest, int tag, MPI_Com
   send(space, sv, dest, tag, comm, DefaultCommMode{});
 }
 
+/// NOTE: This overload has the side effect of fencing on the default execution space.
 template <KokkosView SendView>
 void send(const SendView &sv, int dest, int tag, MPI_Comm comm) {
   send(Kokkos::DefaultExecutionSpace(), sv, dest, tag, comm, DefaultCommMode{});
--- a/src/KokkosComm/mpi/send.hpp
+++ b/src/KokkosComm/mpi/send.hpp
@@ -35,6 +35,7 @@ void send(const ExecSpace &space, const SendView &sv, int dest, int tag, MPI_Com
   };
 
   if (is_contiguous(sv)) {
+    space.fence("fence before send");
     mpi_send_fn(data_handle(sv), span(sv), Impl::mpi_type_v<T>);
   } else {
     auto args = Packer::pack(space, sv);
--- a/src/KokkosComm/mpi/send.hpp
+++ b/src/KokkosComm/mpi/send.hpp
@@ -16,60 +16,30 @@
 
 namespace KokkosComm::mpi {
 
-template <KokkosView SendView, CommunicationMode SendMode>
-void send(const SendView &sv, int dest, int tag, MPI_Comm comm, SendMode) {
-  Kokkos::Tools::pushRegion("KokkosComm::Impl::send");
-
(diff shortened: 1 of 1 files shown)


src/KokkosComm/mpi/send.hpp
```diff
@@ -29,60 +29,30 @@
...
     }
   };
 
-  if (KokkosComm::is_contiguous(sv)) {
-    using SendScalar = typename SendView::value_type;
-    space.fence("fence before MPI_Send");
-    mpi_send_fn(KokkosComm::data_handle(sv), KokkosComm::span(sv), KokkosComm::Impl::mpi_type_v<SendScalar>, dest, tag,
-                comm);
+  if (is_contiguous(sv)) {
+    mpi_send_fn(data_handle(sv), span(sv), Impl::mpi_type_v<T>);
```
cedricchevalier19, line 39: Why `fence()` is not used here?

It looks inconsistent with the previous implementation and the other branch.

cedricchevalier19 (changes requested):
Check for `fence()` consistency.

src/KokkosComm/mpi/send.hpp
dssgabriel: On the non-contiguous path, we fence after the view packing because this may allocate some data and we need to ensure that the view is ready before calling MPI with it.

Because we don't have the packing operation in the contiguous branch I don't think the fence was necessary, hence I removed it.

cedricchevalier19: I think other part of KokkosComm fences because it is kind of expected that you are viewed as added to the queue of the execution space. We fence to ensure that the buffer is ready to be used in a MPI context.

Note that it is on this kind of call that NCCL is interesting, being simply added to the execution queue.

dssgabriel: Yes, this also goes back to what we previously discussed in #108. I am putting the `fence()` back. :+1:

src/KokkosComm/mpi/send.hpp
```diff
@@ -93,4 +64,9 @@ void send(const ExecSpace &space, const SendView &sv, int dest, int tag, MPI_Com
   send(space, sv, dest, tag, comm, DefaultCommMode{});
 }
 
+template <KokkosView SendView>
+void send(const SendView &sv, int dest, int tag, MPI_Comm comm) {
```
cedricchevalier19, line 56: Add a comment that a side effect is to perform a fence on the default execution space.

cedricchevalier19 (dismissed):
Thanks!

dssgabriel:
@cwpearson It looks like you need to add back the SNL label on this PR so it can run on the SNL CI?

cedricchevalier19 (approved):

PR #178, opened 2025-10-09
refactor!: delete `barrier` from high-level API

dssgabriel:
As discussed during the [October 7 developer meeting](https://github.com/kokkos/kokkos-comm/wiki/2025-10-07-meeting-minutes), this PR deletes `barrier` from the high-level API of KokkosComm.

The MPI-specific implementation remains and is available through `KokkosComm::mpi::barrier`, as reflected by the updated unit test.

### DIFF
--- a/src/KokkosComm/collective.hpp
+++ b/src/KokkosComm/collective.hpp
@@ -10,12 +10,4 @@
 #include "fwd.hpp"
 #include "concepts.hpp"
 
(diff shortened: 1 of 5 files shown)


cedricchevalier19 (dismissed):
Make sense, thanks.

cedricchevalier19 (approved):

PR #179, opened 2025-10-24
ci: run NCCL unit tests on SNL CI

dssgabriel:
### Description

This PR enables running the NCCL backend unit tests on the SNL H100 CI workflow.

Some notes:
- NCCL is built from source, packaged into an OS-agnostic tarball & installed from that tarball;
- Kokkos version is explicitly set to 4.7.01 (equivalent to master, aka the latest stable release, at the time of this commit). In the future, we hope to have a container that installs the latest versions of Kokkos/Open MPI/MPICH/NCCL and have it cached by GitHub Actions for reuse in subsequent CI jobs. Submitted jobs will just pull down this container image with all dependencies set up, and only build KokkosComm;
- For now, KokkosComm's MPI and NCCL backends are configured/built/tested one after the other. This may be improved later on, as both are independent and could theoretically run concurrently.

### DIFF
--- a/.github/workflows/snl-h100.yaml
+++ b/.github/workflows/snl-h100.yaml
@@ -1,6 +1,5 @@
 name: H100
 
-
 permissions:
   contents: none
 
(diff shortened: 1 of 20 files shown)


cwpearson (commented):
Fixed up a few things. I'm seeing invalid datatype for `Kokkos::complex<float>` in MPI, but I'm not sure this PR is responsible. @dssgabriel thoughts?

dssgabriel:
@cwpearson NCCL tests are passing on our system :tada: 

Node configuration is:
- 4x GH200 w/ quad-rail Eviden BXI v2 interconnect
- GCC 12.3.0
- CUDA 12.4
- Open MPI 5.0.8
- NCCL 2.21.5

I am still investigating the MPI tests failing on the `complex<T>` dtypes, though.

cwpearson:
@dssgabriel and I identified a few more problems:
* Tough to use a non-blocking communicator for a few reasons
* NCCL packer was using the non-contiguous `span` rather than `size` to compute the count
* We weren't using device-specific execution space instances

dssgabriel (commented):
Looks good to me!

We will need to properly check error codes returned by all NCCL calls in the backend. I will open an issue for tracking that.

We should also consider implementing this for the MPI backend and develop a more robust solution for proper error handling.

cwpearson (approved):

PR #180, opened 2025-10-25
refactor!: build system rework

dssgabriel:
### Description

This PR is a significant rework of the CMake build system, primarily in preparation for a follow-up PR that will introduce CMake presets to streamline the development workflow.

### Exhaustive changes list
- **Bump CMake minimum version to 3.25** (in preparation for CMake workflow presets)
- **Make MPI a required dependency by default**
  Since the NCCL backend also requires MPI for setup, we must require MPI by default
- **Removed all-caps CMake variables overloading existing CMake options with different capitalization**
  Align with CMake best practices and clearly separate usage:
  - Capitalized `KokkosComm_` variables are reserved for CMake
  - All-caps `KOKKOSCOMM_` variables are reserved for source code and pre-processor
- **Renamed options that _build_ additional targets to use `BUILD_` in their name**
- **Ensure that all options _enabling_ a feature/behavior use `ENABLE_` in their name**
- **Move the `KokkosComm` library target declaration and `install` command to the top-level CML**
  - Also removed named file sets since KC is header-only, thus all headers must be installed regardless
- **Reworked packaging logic to align with CMake best practices:**
  - Enable installing the CMake config-file package using an option
    - `ON` if the project is built as top-level, `OFF` for projects consuming KokkosComm 
  - Removed the configure-time CMake-generated `KokkosComm_config.hpp` header
    - Clearer separation of concerns: CMake defines all the necessary variables at configure-time; this is not delegated to a generated header that must be included everywhere
    - Let CMake define all necessary PP variables instead
  - Enforce `ExactVersion` compatibility instead of `SameMajorVersion`
    - Since KC is still in development and pre-v1.0, the API is not stable from version to version
- **Removed all compile flags applied to the `KokkosComm` target**
  - Rationale: because KokkosComm is header-only, it is an `INTERFACE` library in CMake jargon. This means that any targets that link against KokkosComm would inherit the compile flags/definitions/features/etc that we had set here. This is often undesirable and considered bad practice.
  - Note: since KokkosComm is a C++20 project and relies on some of its additions, we still enforce the `cxx_std_20` feature (w/o GNU extensions) so that targets linking against KokkosComm inherit it.
- **Update docs with new CMake options**
- **Update project description & bump KokkosComm version to 0.3.0 to reflect:**
  - This PR's changes
  - Addition of NCCL support (#128 & #179)
  - Misc MPI fixes (#175, #177 & #178)


.github/workflows/linux.yaml
```diff
@@ -23,10 +23,10 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-      - run: echo "The job was automatically triggered by a ${{ github.event_name }} event."
-      - run: echo "This job is now running on a ${{ runner.os }} server hosted by GitHub!"
-      - run: echo "The name of your branch is ${{ github.ref }} and your repository is ${{ github.repository }}."
-      - name: Install MPI
+      - name: Check out repository code
+        uses: actions/checkout@v4
```
cedricchevalier19: LinuxFoundation advise to use commit hash in order to not blindly trust a tag that can be changed any time.

Can be fixed later.

CMakeLists.txt
```diff
@@ -1,92 +1,92 @@
...
-option(KokkosComm_ABORT_ON_ERROR "Runtime error checks trigger a global abort" OFF)
-
-# Resolve options
-set(KOKKOSCOMM_ENABLE_PERFTESTS ${KokkosComm_ENABLE_PERFTESTS} CACHE BOOL "" FORCE)
-set(KOKKOSCOMM_ENABLE_TESTS ${KokkosComm_ENABLE_TESTS} CACHE BOOL "" FORCE)
-set(KOKKOSCOMM_ENABLE_MPI ${KokkosComm_ENABLE_MPI} CACHE BOOL "" FORCE)
-set(KOKKOSCOMM_ENABLE_NCCL ${KokkosComm_ENABLE_NCCL} CACHE BOOL "" FORCE)
+# --- OPTIONS --- #
+option(KokkosComm_BUILD_UNITTESTS "Build unit tests. Default: OFF. Values: { ON, OFF }." OFF)
+option(KokkosComm_BUILD_PERFTESTS "Build perf tests. Default: OFF. Values: { ON, OFF }." OFF)
```
cedricchevalier19:
```suggestion
option(KokkosComm_BUILD_UNITTESTS "Build unit tests." OFF)
option(KokkosComm_BUILD_PERFTESTS "Build perf tests. " OFF)
```

options are boolean and ccmake or these kind of tools are smart enough to print the default values. It will be easier to maintain to not duplicate this information.

CMakeLists.txt
```diff
@@ -1,92 +1,92 @@
...
+  "Install a CMake config-file package. Default: ${PROJECT_IS_TOP_LEVEL}. Values: { ON, OFF }."
+  ${PROJECT_IS_TOP_LEVEL}
+)
+option(KokkosComm_ENABLE_MPI "Enable MPI communication space backend. Default: ON. Values: { ON, OFF }." ON)
+option(KokkosComm_ENABLE_NCCL "Enable NCCL communication space backend. Default: OFF. Values: { ON, OFF }." OFF)
+option(
+  KokkosComm_ENABLE_ABORT_ON_ERROR
+  "Enable global abort on runtime error checks. Default: OFF. Values: { ON, OFF }."
+  OFF
+)
```
cedricchevalier19: Same here.

CMakeLists.txt
```diff
@@ -1,92 +1,92 @@
...
-set(KOKKOSCOMM_ENABLE_MPI ${KokkosComm_ENABLE_MPI} CACHE BOOL "" FORCE)
-set(KOKKOSCOMM_ENABLE_NCCL ${KokkosComm_ENABLE_NCCL} CACHE BOOL "" FORCE)
+# --- OPTIONS --- #
+option(KokkosComm_BUILD_UNITTESTS "Build unit tests. Default: OFF. Values: { ON, OFF }." OFF)
+option(KokkosComm_BUILD_PERFTESTS "Build perf tests. Default: OFF. Values: { ON, OFF }." OFF)
+option(
+  KokkosComm_INSTALL_CMAKE_PACKAGE
+  "Install a CMake config-file package. Default: ${PROJECT_IS_TOP_LEVEL}. Values: { ON, OFF }."
+  ${PROJECT_IS_TOP_LEVEL}
+)
```
cedricchevalier19: Can you describe a use case for that?
I can understand why one can add Kokkos_Comm directly into their project to ease development but do we want them to install their software like this?

CMakeLists.txt
```diff
@@ -1,92 +1,92 @@
...
+  "Enable global abort on runtime error checks. Default: OFF. Values: { ON, OFF }."
+  OFF
+)
 
+# --- DEPENDENCIES --- #
 find_package(Kokkos REQUIRED)
-if(KOKKOSCOMM_ENABLE_MPI)
-  find_package(MPI REQUIRED)
-endif()
+find_package(MPI REQUIRED)
```
cedricchevalier19: Explain why MPI is mandatory.

I would rather have a duplicated `find_package(MPI REQUIRED)` for tests or NCCL if they are the one requiring it.

CMakeLists.txt
```diff
@@ -1,92 +1,92 @@
...
-# Generate package version file
-write_basic_package_version_file(
-  ${PROJECT_BINARY_DIR}/cmake/KokkosComm/KokkosCommVersion.cmake
-  COMPATIBILITY SameMajorVersion
-  ARCH_INDEPENDENT # KokkosComm is a header-only library
-)
+  # Generate package version-file
+  write_basic_package_version_file(
+    ${PROJECT_BINARY_DIR}/KokkosCommVersion.cmake
+    COMPATIBILITY ExactVersion
```
cedricchevalier19, line 61: Why this change?

src/KokkosComm/CMakeLists.txt
```diff
@@ -101,97 +71,46 @@ if(KOKKOSCOMM_ENABLE_NCCL)
...
+  if(KokkosComm_IMPL_MPI_HAS_MPI_EXT_H)
+    target_compile_definitions(KokkosComm INTERFACE KOKKOSCOMM_IMPL_MPIEXT_H)
+  endif()
+endif()
+if(KokkosComm_ENABLE_NCCL)
+  target_compile_definitions(KokkosComm INTERFACE KOKKOSCOMM_ENABLE_NCCL)
+endif()
+if(KokkosComm_ENABLE_ABORT_ON_ERROR)
+  target_compile_definitions(KokkosComm INTERFACE KOKKOSCOMM_ABORT_ON_ERROR)
+endif()
```
cedricchevalier19, line 107: Just a personal opinion, but I think having a `KokkosCommConfig.h`file with all this logic is easier to follow. It can also help to check if the configuration is correct, without having to compile something.

I think having `KokkosComm_IMPL_MPI_HAS_MPI_EXT_H` defined from CMake is better than the test in the header file we previously had.

unit_tests/test_main.cpp
```diff
@@ -74,8 +73,7 @@ int main(int argc, char *argv[]) {
   MPI_Comm_rank(MPI_COMM_WORLD, &rank);
   MPI_Comm_size(MPI_COMM_WORLD, &size);
   if (0 == rank) {
-    std::cerr << argv[0] << " (KokkosComm " << KOKKOSCOMM_VERSION_MAJOR << "." << KOKKOSCOMM_VERSION_MINOR << "."
-              << KOKKOSCOMM_VERSION_PATCH << ")\n";
+    std::cerr << argv[0] << "\n";
```
cedricchevalier19, line 76: It is convenient to have the version of KokkosComm printed when you want to monitor a CI. Ideally having the git hash can be great too!

cmake/mpi-vendor.cmake
cedricchevalier19, line 1: On the long run we do not wish to keep this distinction between MPI libraries, except of specific backends.

We can leave it here for this PR but I hope we can get rid of it later, perhaps you can add a FIXME at the top of the file.

cedricchevalier19 (commented):
Thanks Gabriel, I think it is quite an improvement.
 
If you have some arguments for removing the header config file, I am listening :).

.github/workflows/linux.yaml
dssgabriel: Fixed in #4f40318

CMakeLists.txt
dssgabriel: Fixed in #0d36ccd

CMakeLists.txt
dssgabriel: Fixed in #0d36ccd

CMakeLists.txt
dssgabriel: This is aimed at improving the workflow for users who install KokkosComm on their system (i.e., what we test in the `linux-install` CI workflow).

Producing a CMake config-file package and installing it in a directory that is in `$PATH` will let CMake's `find_package()` automatically find KokkosComm without having to pass `-DKokkosComm_ROOT` on the CMake configuration command.
Additionally, this is the recommended, standard way of packaging libraries in modern CMake (see, e.g., [this talk from a CMake developer](https://youtu.be/K5Kg8TOTKjU)).

CMakeLists.txt
dssgabriel: Pondering on this, I think you are right. MPI is only really required if tests are enabled, so I think this should be reverted to being behind the `KokkosComm_ENABLE_MPI` check.

dssgabriel: Done in #6df4962.

CMakeLists.txt
dssgabriel: This function generates (in part) the CMake config-file package. The `COMPATIBILITY` parameter controls the version compatibility check when CMake calls `find_package`.
Since KokkosComm is still in active development and pre-v1.0, the API is not stable from version to version and may introduce breaking changes; hence, the change to an exact version compatibility requirement.

unit_tests/test_main.cpp
dssgabriel: This can be retrieved through CMake and printed in the CI output, so I am unsure why it is necessary to have it available in the executable? Reliable and robust library packaging using CMake config-file packages is the current best solution for exposing such information, IMHO.

cmake/mpi-vendor.cmake
dssgabriel: Addressed in #f9739e8

dssgabriel:
I am not sure I understand why the MPICH-Debug job fails...
The failing test is `mpi/test_sendrecv`: all contiguous tests are skipped because there is only one rank (but the program is correctly launched with 2 processes, so that's already weird), and non-contiguous tests directly fail (probably because we don't check that there are two ranks).

I do not understand why there is only one rank since we run the test with:
```sh
/usr/bin/mpiexec "-n" "2" "./test-sendrecv"
```
And why do all other tests pass correctly, `test_isendirecv` included?

unit_tests/test_main.cpp
cedricchevalier19: I think it is just more convenient to have just one log file to read. It also allows to test if these macros are properly defined.

dssgabriel:
Closing this in favor of #212

PR #181, opened 2025-11-06
Allow Views of rank > 1 in collective communication operations

nicoleavans:
The semantics used by Tpetra within Trilinos conflict with the limitation that `SendView::rank > 1 not supported` / `RecvView::rank > 1 not supported`. I have removed the two within `src/KokkosComm/mpi/allreduce.hpp` for my own experimentation replacing MPI with Kokkos Comm in Tpetra. I have also removed the equivalent lines from:

- `src/KokkosComm/mpi/allgather.hpp`
- `src/KokkosComm/mpi/alltoall.hpp`

This same style has proliferated through the NCCL communication backend (e.g., the corresponding NCCL [`allgather`](https://github.com/kokkos/kokkos-comm/blob/f9a01894a61550b126f51a9b63b6780907e9559b/src/KokkosComm/nccl/allgather.hpp#L26)). A consensus on the style going forward should be reached. I support removing this limitation.

### DIFF
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -89,6 +89,11 @@ Receive
 Collectives
 ===========
 
+.. important:: 
+    Collective operations act **element-wise** on the input Views. Multi-dimensional Views are treated as a **logically flattened** sequence of values, and the reduction is applied over that sequence. All participating Views must have **identical extents**; mismatched shapes result in undefined behavior.
+
+    The reduction operator must be **associative**, but ordering of partial combinations is **not guaranteed**, and the operation is not required to be commutative.
+
 .. cpp:namespace:: KokkosComm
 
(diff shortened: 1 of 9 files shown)


dssgabriel:
As mentioned during today's meeting, I am in favor of lifting the multi-dimensional constraints on collective operations :slightly_smiling_face:
This can be done for both the MPI and NCCL backends; however, we need well-defined semantics for multi-dimensional views, which should be reflected in our documentation.

This will be a very welcome change that broadens the capabilities of KokkosComm!

nicoleavans:
Still need to update docs

nicoleavans:
After re-reviewing the documentation, I don't think any changes need to be made within the scope of this PR. I could find no mention of this restriction. I believe it is ready for review. @cwpearson or @dssgabriel could you please add reviewers? I don't have permissions to do it myself, for some reason.

cedricchevalier19:
I think it can be clarified a little bit. Like that the operation is element wise and that the Views have to have the same shape (we cannot check this at compile time).
It might look natural for a MPI point of view, but from a Kokkos user, operations like reductions are performed according to the data, not to threads or processors.

cedricchevalier19 (approved):

PR #183, opened 2025-11-06
fix!: make "core" `broadcast` in-place only

### LINKED ISSUE #182: `Broadcast` struct cannot be specialized for MPI
The `Broadcast` struct introduced in #128
https://github.com/kokkos/kokkos-comm/blob/f9a01894a61550b126f51a9b63b6780907e9559b/src/KokkosComm/fwd.hpp#L50-L52
cannot be specialized for the MPI backend because the C MPI API only supports one buffer, i.e. it is always in-place. This is a departure from what NCCL supports, as they have both a send and receive buffer:
```cpp
int ncclBroadcast(void* sendbuff, void* recvbuff, size_t count, ncclDatatype_t, datatype, int root, ncclComm_t comm, cudaStream_t stream);
```
Additionally, they support a "legacy" MPI-like API for doing an in-place broadcast:
```cpp
int ncclBcast(void* buff, size_t count, ncclDatatype_t, datatype, int root, ncclComm_t comm, cudaStream_t stream);
```
Which is equivalent to the first one with `sendbuff == recvbuff`.

In order to be able to bind the MPI backend to our "core" API, we need to change the `Broadcast` struct to only have a single view, and fix the NCCL backend accordingly (thus only supporting "in-place" broadcast).

dssgabriel:
This PR fixes #182 in order to match MPI's C API and enable binding the MPI backend to the "core" API of KokkosComm (in a follow-up PR).
This change also forces the NCCL backend to only expose the in-place version of its broadcast (which was the only one we tested anyway).

---
**Sidenote:** Although this technically restricts the NCCL backend, I find it to be a positive change anyway, as it semantically means that the passed view is always used, regardless of the process's rank.
In the NCCL API, `sv` is ignored by all ranks apart from the root, and conversely, `rv` is ignored on the root rank:
```cpp
ncclBroadcast(sbuf, rbuf, count, dtype, root, comm, stream);
```
IMHO, having a different "meaning" for the passed views depending on the rank wasn't very elegant.

### DIFF
--- a/src/KokkosComm/collective.hpp
+++ b/src/KokkosComm/collective.hpp
@@ -22,20 +22,11 @@
 
 namespace KokkosComm::Experimental {
 
-/// Copy the `sv` view on the `root` rank to all ranks' `rv` view.
-///
-/// The `sv` view is only used on the `root` rank and ignored for all other ranks.
-template <KokkosView SendView, KokkosView RecvView, KokkosExecutionSpace ExecSpace = Kokkos::DefaultExecutionSpace,
-          CommunicationSpace CommSpace = DefaultCommunicationSpace>
-auto broadcast(Handle<ExecSpace, CommSpace>& h, const SendView sv, RecvView rv, int root) -> Req<CommSpace> {
-  return Impl::Broadcast<SendView, RecvView, ExecSpace, CommSpace>::execute(h, sv, rv, root);
-}
-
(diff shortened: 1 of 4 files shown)


cedricchevalier19 (approved):
It makes sense to have a simple API.

dssgabriel:
@cwpearson Could you please add the SNL-CI label so this could go through that workflow before we merge?

PR #186, opened 2025-11-26
refactor(mpi): move `mpi_type` in the `mpi` namespace

dssgabriel:
This places the `mpi_type` datatype conversion function (and friends) in the `mpi` namespace. The remaining changes are intended to accommodate any code that would otherwise fail to compile.

In the future, we may consider rethinking how this is implemented.
E.g., we could expose a generic function, templated on the communication space and specialized as necessary:
```cpp
template <CommunicationSpace CS, typename T>
[[no_discard]] constexpr auto dtype() -> CS::datatype_type;

static_assert(dtype<Mpi, double>() == MPI_DOUBLE);
```


dssgabriel:
Closed in favor of #189

PR #187, opened 2025-11-26
feat: add reduction ops conversion for MPI + cleanup impl

dssgabriel:
- Make reduction operator conversion `CommSpace`-generic
  - Perform necessary changes for NCCL
- Added conversion logic between KokkosComm's reduction operators and `MPI_Op`
  - Added missing `LXor` and `BXor` operators
- Cleaned up primary impl header
  - updated (missed) copyright header
  - fix includes (IWYU)
  - removed unnecessary namespace prefixing

### DIFF
--- a/src/KokkosComm/reduction_op.hpp
+++ b/src/KokkosComm/reduction_op.hpp
@@ -21,7 +21,7 @@ namespace {
 
 // clang-format off
 #define DECL_REDUCTION_OP_FOR(operator) \
-  struct operator;                      \
+  struct operator {};                   \
   template <> struct Impl::is_reduction_operator<operator> : public std::true_type {}
 // clang-format on
 
--- a/src/KokkosComm/reduction_op.hpp
+++ b/src/KokkosComm/reduction_op.hpp
@@ -1,20 +1,9 @@
-//@HEADER
-// ************************************************************************
-//
-//                        Kokkos v. 4.0
-//       Copyright (2022) National Technology & Engineering
-//               Solutions of Sandia, LLC (NTESS).
-//
-// Under the terms of Contract DE-NA0003525 with NTESS,
-// the U.S. Government retains certain rights in this software.
-//
-// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
-// See https://kokkos.org/LICENSE for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-//
-//@HEADER
+// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
 
 #pragma once
+
+#include <type_traits>
 
 #include <Kokkos_Core.hpp>
 #ifdef KOKKOSCOMM_ENABLE_NCCL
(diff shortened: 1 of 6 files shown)


src/KokkosComm/nccl/impl/reduction_op.hpp
```diff
@@ -0,0 +1,36 @@
...
+constexpr auto reduction_op() -> ncclRedOp_t {
+  if constexpr (std::is_same_v<RedOp, KokkosComm::Max>) {
+    return ncclMax;
+  } else if constexpr (std::is_same_v<RedOp, KokkosComm::Min>) {
+    return ncclMin;
+  } else if constexpr (std::is_same_v<RedOp, KokkosComm::Sum>) {
+    return ncclSum;
+  } else if constexpr (std::is_same_v<RedOp, KokkosComm::Prod>) {
+    return ncclProd;
+  } else if constexpr (std::is_same_v<RedOp, KokkosComm::Average>) {
```
cedricchevalier19: Why using namespace here and not for MPI?

src/KokkosComm/reduction_op.hpp
```diff
@@ -1,107 +1,72 @@
...
+#include <type_traits>
 
 #include <KokkosComm/concepts.hpp>
 
 namespace KokkosComm {
 
 struct BAnd {};
 template <>
-struct KokkosComm::Impl::is_reduction_operator<KokkosComm::BAnd> : public std::true_type {};
+struct Impl::is_reduction_operator<BAnd> : public std::true_type {};
```
cedricchevalier19: I think a macro may be useful here to declare and add the `is_reduction_operator`.

dssgabriel: Done :+1:

cedricchevalier19 (dismissed):
LGTM

cedricchevalier19 (commented):
I am not sure how to test the dispatch logic, perhaps adding a unit test that almost looks like your implementation but that can save us later when changing the reduction op dispatch.

Can you use these reduction op in MPI, in this PR?

Can we make something to be able to know if a reduction operator is implemented for a given `CommunicationSpace`?

dssgabriel:
> perhaps adding a unit test that almost looks like your implementation, but that can save us later when changing the reduction op dispatch

Yes, definitely. I will add conversion tests in this PR.

---
> Can you use these reduction op in MPI, in this PR?

Do you mean for the existing low-level MPI reduction interfaces we have? I could, but this would mean that these no longer take an `MPI_Op` parameter, which would not be consistent with what I did for the low-level NCCL bindings (which take an `ncclRedOp_t` parameter instead). As I understand our design logic, low-level bindings (the ones in the `mpi::` or `nccl::` sub-namespaces) use the backend's own data types (e.g. `MPI_Comm`/`ncclComm_t`, etc.).
PRs #197 and #199 are the ones that introduce the use of the generic `KokkosComm::ReductionOperator` concepts for their reduction op parameter, since they are part of the core API (consistent with the NCCL specialization).

---
> Can we make something to be able to know if a reduction operator is implemented for a given CommunicationSpace?

I could write a helper function like:
```cpp
template <CommunicationSpace CS, ReductionOperator RO>
auto is_reduction_op_convertible() -> bool;
```
Is this what you meant?

Additionally, this could become part of the constraints on the `ReductionOperator` concept, which may allow us to produce better diagnostics if a reduction operator isn't available for a given backend.

dssgabriel:
@cedricchevalier19 I added the conversion test for reduction operators, as discussed above.

What would be left to do before merging this PR?

cedricchevalier19 (approved):
Thanks for this work.

cedricchevalier19:
@cwpearson can you run the Sandia CI?

PR #188, opened 2025-11-28
refactor!: add more requirements to the `CommunicationSpace` concept

dssgabriel:
This PR constrains the `CommunicationSpace` concept with the following required member type aliases:
- `typename communication_space`
- `typename handle_type`
- `typename request_type`
- `typename datatype_type`
- `typename reduction_op_type`
- `typename rank_type`

Of course, we may add more/remove some of these in the future.

This allows us to implement `CommunicationSpace`-generic APIs for Kokkos Comm (e.g. the datatype conversion alternative mentioned in #186).

I also renamed our existing communication spaces `Mpi` and `Nccl` to `MpiSpace` and `NcclSpace` respectively, and propagated the change everywhere necessary.
This removes any naming ambiguity, matches the name of the header they are declared in, and also happens to correspond to the naming semantic using in Kokkos Remote Spaces (see, e.g., https://github.com/kokkos/kokkos-remote-spaces/blob/main/src/impl/mpispace/Kokkos_MPISpace.hpp).

### DIFF
--- a/src/KokkosComm/mpi/mpi_space.hpp
+++ b/src/KokkosComm/mpi/mpi_space.hpp
@@ -11,26 +11,17 @@
 
 namespace KokkosComm {
 
-// TODO: not sure what members this thing needs
-struct Mpi {
-  // TODO: just an example
-  static int world_size() {
-    int size;
-    MPI_Comm_size(MPI_COMM_WORLD, &size);
-    return size;
-  }
+struct MpiSpace {
+  using communication_space = MpiSpace;
+  using handle_type         = MPI_Comm;
+  using request_type        = MPI_Request;
+  using datatype_type       = MPI_Datatype;
+  using reduction_op_type   = MPI_Op;
+  using rank_type           = int;  
+};
 
-  // TODO: just an example
-  static int world_rank() {
-    int rank;
-    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
-    return rank;
-  }
-
-};  // struct Mpi
-
-// KokkosComm::Mpi is a KokkosComm::CommunicationSpace
+// KokkosComm::MpiSpace is a KokkosComm::CommunicationSpace
 template <>
-struct Impl::is_communication_space<KokkosComm::Mpi> : public std::true_type {};
+struct Impl::is_communication_space<MpiSpace> : public std::true_type {};
 
 }  // namespace KokkosComm
--- a/src/KokkosComm/mpi/mpi_space.hpp
+++ b/src/KokkosComm/mpi/mpi_space.hpp
@@ -11,13 +11,14 @@
 
 namespace KokkosComm {
 
+/// The MPI communication space.
 struct MpiSpace {
   using communication_space = MpiSpace;
   using handle_type         = MPI_Comm;
   using request_type        = MPI_Request;
   using datatype_type       = MPI_Datatype;
   using reduction_op_type   = MPI_Op;
-  using rank_type           = int;  
+  using rank_type           = int;
 };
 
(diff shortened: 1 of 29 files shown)


src/KokkosComm/mpi/mpi_space.hpp
```diff
@@ -11,26 +11,18 @@
...
-
-// KokkosComm::Mpi is a KokkosComm::CommunicationSpace
+/// The MPI communication space.
+struct MpiSpace {
+  using communication_space = MpiSpace;
+  using handle_type         = MPI_Comm;
+  using request_type        = MPI_Request;
+  using datatype_type       = MPI_Datatype;
+  using reduction_op_type   = MPI_Op;
+  using rank_type           = int;
```
cedricchevalier19, line 21: I am not sure it is a good idea to have this public.

src/KokkosComm/nccl/broadcast.hpp
```diff
@@ -18,7 +18,8 @@ namespace nccl {
 namespace KC = KokkosComm;
 
 template <KokkosExecutionSpace ExecSpace, KokkosView SendView, KokkosView RecvView>
-auto broadcast(const ExecSpace &space, const SendView &sv, const RecvView &rv, int root, ncclComm_t comm) -> Req<Nccl> {
+auto broadcast(const ExecSpace &space, const SendView &sv, const RecvView &rv, int root, ncclComm_t comm)
```
cedricchevalier19: Does this works with any kind of `ExecSpace`?

src/KokkosComm/nccl/handle.hpp
```diff
@@ -13,17 +13,17 @@
 namespace KokkosComm {
 
 template <>
-class Handle<Kokkos::Cuda, Experimental::Nccl> {
+class Handle<Kokkos::Cuda, Experimental::NcclSpace> {
```
cedricchevalier19, line 16: Not for this PR, but we should provide `Handle` for any kind of `Cuda*Space`.

src/KokkosComm/mpi/mpi_space.hpp
dssgabriel: Why wouldn't it be? There is nothing special about these typename definitions, and it would be preferable for users to use these, rather than have explicit communication space-specific data types mixed with Kokkos Comm code.

src/KokkosComm/nccl/broadcast.hpp
dssgabriel: I left it like this because specializing it to `Kokkos::Cuda` would prevent us from reusing this low-level interface for interoperability with AMD RCCL.

src/KokkosComm/mpi/mpi_space.hpp
cedricchevalier19: My question was more: do we commit to have these `using` on the long run.

> There is nothing special about these typename definitions, and it would be preferable for users to use these, 

It is not that the code is complicated, but that we will not be able to modify easily what we store here. Are we committed to `handle_type`, ...
If these fields are not used publicly yet, I think we can discuss a little bit more on accessibility.

> rather than have explicit communication space-specific data types mixed with Kokkos Comm code.

They are for `MpiSpace` other communication spaces might define something else so users will have communication space-specific stuff. With your `using` we provide a way for them to use the correct types, but I need to think more on the impact of providing these fields.

cedricchevalier19: I'll merge this PR as is and we will review the accessibility when dealing with #200 .

cedricchevalier19 (approved):

PR #190, opened 2025-12-01
feat(mpi): add non-blocking broadcast

dssgabriel:
Add `mpi::ibroadcast` in preparation for exposure in "core" API.

Based on #183, so it needs to be merged first.

PR #191, opened 2025-12-01
feat(core): expose MPI-based broadcast

dssgabriel:
Expose the MPI-based broadcast in the "core" API.
Also, add the associated unit test, which now works with either the MPI or NCCL communication spaces.

Based on #190, which is itself based on #183, so both PRs need to be merged first.


cedricchevalier19 (commented):
We should add this function in the documentation.

cedricchevalier19 (approved):

cedricchevalier19:
Should update the documentation #201

PR #192, opened 2025-12-01
feat(mpi): add non-blocking all-gather

dssgabriel:
Prepare for exposure in "core" API.


cedricchevalier19 (commented):
Can we test this internal function?

dssgabriel:
> Can we test this internal function?

Internal function will be transitively tested by the "core" all-gather test added in #193.

cedricchevalier19 (approved):

PR #193, opened 2025-12-01
feat(core): expose MPI-based all-gather

dssgabriel:
Also, add the associated unit test.

Based on #192, which will need to be merged first.


cedricchevalier19 (dismissed):
Looks ok.

cedricchevalier19 (commented):
We should add this in the documentation.

cedricchevalier19 (approved):
I am ok to merge and refactor the documentation later (#201).

PR #196, opened 2025-12-02
build(unit_tests): custom CMake helper for adding unit tests

dssgabriel:
This PR adds a custom CMake helper function for easily adding new unit tests.

I made it as flexible as possible, so it should cover all cases we may encounter for existing and future unit tests. Also added some custom `gersemi` definitions so our CMake formatter can parse and format these function calls properly.

Applied this new function to all our existing unit tests and renamed them using the following format: `$kind.$module.$name`, where:
- kind: `{smoke,test}`
- module: `{core,mpi,nccl}`

Example call for the pending PR #191 that adds the "core" `broadcast` unit test:
```cmake
kc_add_unit_test(
  test.core.broadcast                     # Full test name
  CORE                                    # Test module
  NUM_PES 2                               # Number of Processing Elements (PEs) to run the test with
  FILES test_main.cpp test_broadcast.cpp  # Files defining the test
)
```

### DIFF
--- a/cmake/kc-test.cmake
+++ b/cmake/kc-test.cmake
@@ -23,5 +23,5 @@ function(kc_add_unit_test name)
     target_link_libraries(${name} PRIVATE NCCL::NCCL)
   endif()
 
-  add_test(NAME ${name} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${UT_NUM_PES} ${name})
+  add_test(NAME ${name} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${UT_NUM_PES} ./${name})
 endfunction()
--- a/.cmake-format
+++ b/.cmake-format
@@ -1,4 +1,4 @@
-definitions: [./cmake/kokkoscomm_test.cmake]
+definitions: [./cmake/kc-test.cmake]
 line_length: 120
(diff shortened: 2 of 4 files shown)


cmake/kc-test.cmake
```diff
@@ -0,0 +1,27 @@
...
+  cmake_parse_arguments(UT "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+  add_executable(${name})
+  target_sources(${name} PRIVATE ${UT_FILES})
+  target_compile_features(${name} PRIVATE cxx_std_20)
+  target_compile_options(${name} PRIVATE ${UT_OPTIONS})
+  target_include_directories(${name} PRIVATE ${UT_INCLUDES})
+  target_link_libraries(${name} PRIVATE gtest fmt::fmt ${UT_LIBRARIES})
+
+  if(UT_CORE)
```
cedricchevalier19, line 15: What is `CORE`?

dssgabriel: `CORE` indicates that the test is for the "core", high-level API, i.e. one of `KokkosComm::{send,recv,broadcast,allgather,alltoall,allreduce,reduce}`.

Having these parameters (`CORE`, `MPI`, `NCCL`) makes it easier to know which libraries need to be linked by default, which would otherwise be annoying to do properly in the unit tests CML.

cedricchevalier19 (approved):

PR #202, opened 2025-12-16
Disable file reporter from other processes

yasahi-hpc:
This PR aims at disabling file output from other processes.
Without this, the json output will be contaminated.


dssgabriel (approved):
LGTM

cedricchevalier19 (approved):
Thanks for the fix @yasahi-hpc.

I think we should follow up later, as we may want to generate (different) output for each rank.

yasahi-hpc:
> Thanks for the fix @yasahi-hpc.
> 
> I think we should follow up later, as we may want to generate (different) output for each rank.

Thank you for merging.
Yeah, that could be a good option

PR #198, opened 2025-12-04
feat(mpi): add non-blocking reduce (`mpi:ireduce`)

dssgabriel:
Prepare for exposure in "core" API.

Based on #187 and #188.

PR #208, opened 2026-01-20
feat(mpi): implement host staging

dssgabriel:
This PR is based on #205, using the proposed host staging API.
It enables automatic host staging for the MPI backend, when the provided MPI implementation is not GPU-aware (controlled via a CMake option defined at config-time: `-DKokkosComm_ENABLE_GPU_AWARE_MPI=ON`).

**To-Do list of interfaces to cover**

P2P:
- [x] `mpi::send`
- [x] `mpi::isend`
- [x] `mpi::recv`
- [x] `mpi::irecv`

Colls:
- [x] `mpi::broadcast`
- [x] `mpi::ibroadcast`
- [x] `mpi::allgather`
- [x] `mpi::iallgather`
- [ ] `mpi::allreduce`
- [ ] `mpi::iallreduce`
- [ ] `mpi::reduce`
- [ ] `mpi::ireduce` (depends on #198)
- [x] `mpi::alltoall`
- [x] `mpi::ialltoall`
- [ ] `mpi::inclusive_scan`
- [ ] `mpi::exclusive_scan`

Some questions/notes about host staging implementation:
- For non-blocking "send-only" interfaces, do we need to extend the lifetime of the passed view? Technically, only the host-staged view needs to have its lifetime extended until the completion of the communication, not the provided view. E.g.:
  ```cpp
  auto host_sv = KokkosComm::Impl::stage_for(sv);
  h.space().fence("fence host staging before `MPI_Isend`");
  
  // Assume contiguous view
  MPI_Isend(data_handle(host_sv), span(host_sv), datatype<MpiSpace, T>(), dest, tag, h.mpi_comm(), &req.mpi_request());
  req.extend_view_lifetime(host_sv);

  // NOTE: Is this really needed?
  req.extend_view_lifetime(sv);
  ```
- For non-contiguous "receive" interfaces, can we directly unpack into the passed view instead of the host-staged view? This would remove a call to `deep_copy`, which I think is smart enough to do the right thing, but I am not sure. E.g.:
  ```cpp
  auto host_rv = KokkosComm::Impl::stage_for(rv);
  space.fence("fence host staging before `MPI_Recv`");
  
  // Assume non-contiguous view 
  auto packed = Packer::allocate_packed_for(space, "packed `MPI_Recv`", host_rv);
  space.fence("fence packing before `MPI_Recv`");
  MPI_Recv(data_handle(packed.view), packed.count, packed.datatype, src, tag, comm, MPI_STATUS_IGNORE);

  // NOTE: Can we unpack directly into `rv` instead of `host_rv`
  // and eliminate the subsequent call to `copy_back`?
  Packer::unpack_into(space, host_rv, args.view);
  KokkosComm::Impl::copy_back(space, rv, host_rv);

  space.fence("fence copy back after `MPI_Recv`");
  ```

### DIFF
--- a/src/KokkosComm/mpi/allgather.hpp
+++ b/src/KokkosComm/mpi/allgather.hpp
@@ -41,55 +41,48 @@ auto iallgather(const ExecSpace &space, const SView sv, RView rv, MPI_Comm comm)
   return req;
 }
 
-template <KokkosView SendView, KokkosView RecvView>
-void allgather(const SendView &sv, const RecvView &rv, MPI_Comm comm) {
-  Kokkos::Tools::pushRegion("KokkosComm::Mpi::allgather");
-
-  using SendScalar = typename SendView::value_type;
-  using RecvScalar = typename RecvView::value_type;
-
-  static_assert(KokkosComm::rank<SendView>() <= 1, "allgather for SendView::rank > 1 not supported");
-  static_assert(KokkosComm::rank<RecvView>() <= 1, "allgather for RecvView::rank > 1 not supported");
-
-  KokkosComm::mpi::fail_if(!KokkosComm::is_contiguous(sv), "low-level allgather requires contiguous send view");
-  KokkosComm::mpi::fail_if(!KokkosComm::is_contiguous(rv), "low-level allgather requires contiguous recv view");
-
-  const int count = KokkosComm::span(sv);  // all ranks send/recv same count
-  MPI_Allgather(KokkosComm::data_handle(sv), count, datatype<MpiSpace, SendScalar>(), KokkosComm::data_handle(rv),
-                count, datatype<MpiSpace, RecvScalar>(), comm);
-
-  Kokkos::Tools::popRegion();
-}
-
-// in-place allgather
-template <KokkosExecutionSpace ExecSpace, KokkosView RecvView>
-void allgather(const ExecSpace &space, const RecvView &rv, const size_t recvCount, MPI_Comm comm) {
-  Kokkos::Tools::pushRegion("KokkosComm::Mpi::allgather");
-
-  using RecvScalar = typename RecvView::value_type;
-
-  static_assert(KokkosComm::rank<RecvView>() <= 1, "allgather for RecvView::rank > 1 not supported");
-
-  KokkosComm::mpi::fail_if(!KokkosComm::is_contiguous(rv), "low-level allgather requires contiguous recv view");
-
-  space.fence("fence before allgather");  // work in space may have been used to produce send view data
-  MPI_Allgather(MPI_IN_PLACE, 0 /*ignored*/, MPI_DATATYPE_NULL /*ignored*/, KokkosComm::data_handle(rv), recvCount,
-                datatype<MpiSpace, RecvScalar>(), comm);
-
-  Kokkos::Tools::popRegion();
-}
-
 template <KokkosExecutionSpace ExecSpace, KokkosView SendView, KokkosView RecvView>
-void allgather(const ExecSpace &space, const SendView &sv, const RecvView &rv, MPI_Comm comm) {
-  Kokkos::Tools::pushRegion("KokkosComm::Mpi::allgather");
+void allgather(const ExecSpace& space, const SendView& sv, const RecvView& rv, MPI_Comm comm) {
+  using ST = typename SendView::non_const_value_type;
+  using RT = typename RecvView::non_const_value_type;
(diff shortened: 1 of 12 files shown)


src/KokkosComm/mpi/allgather.hpp
```diff
@@ -12,92 +12,119 @@
...
+#if defined(KOKKOSCOMM_ENABLE_GPU_AWARE_MPI)
+  // Sync: Work in space may have been used to produce view data.
+  space.fence("fence before GPU-aware `MPI_Iallgather`");
+  MPI_Iallgather(data_handle(sv), cnt, datatype<MpiSpace, ST>, data_handle(rv), cnt, datatype<MpiSpace, RT>, comm,
+                 &req.mpi_request());
   req.extend_view_lifetime(sv);
   req.extend_view_lifetime(rv);
+#else
+  auto host_sv = KokkosComm::Impl::stage_for(sv);
+  auto host_rv = KokkosComm::Impl::stage_for(rv);
```
cedricchevalier19, line 43: Not sure if calling `create_mirror_view_and_copy` and `create_mirror_view` explicitly is not better

src/KokkosComm/mpi/allgather.hpp
```diff
@@ -12,92 +12,119 @@
...
+  space.fence("fence host staging before `MPI_Iallgather`");
+  MPI_Iallgather(data_handle(host_sv), cnt, datatype<MpiSpace, ST>, data_handle(host_rv), cnt, datatype<MpiSpace, RT>,
+                 comm, &req.mpi_request());
+  // Implicitly extends lifetimes of `host_rv` and `rv` due to lambda capture
+  req.call_after_mpi_wait([=]() {
+    KokkosComm::Impl::copy_back(space, rv, host_rv);
```
cedricchevalier19, line 49: Check if `copy_back` is needed or if we can directly use `deep_copy` (that should be no-op when the two views are the same).

src/KokkosComm/mpi/allgather.hpp
```diff
@@ -12,92 +12,119 @@
...
+    space.fence("fence copy back after `MPI_Iallgather`");
+  });
+  req.extend_view_lifetime(host_sv);
+  req.extend_view_lifetime(sv);
```
cedricchevalier19, line 53: Why we need this?

cedricchevalier19 (commented):
Just some thoughts.

To me the biggest issue with staging is how we manage the memory (the same goes for packing).

I am not convinced that we have to wrap Kokkos mirror functions or deep_copy.

src/KokkosComm/mpi/allgather.hpp
dssgabriel: We don't in the case of a host-staged send operation, since the view that is actually sent (and that needs to live long enough) is `host_sv`, not `sv`.
This could be safely removed, but our docs should clearly state what the semantics of KC calls are with respect to view reuse.

In the host-staged case, while `sv` is technically reusable by the user immediately after the KC collective is called, I think it would be better to have the same semantics in both execution paths, and to mandate that `sv` is reusable only after the comm operation completes (via `wait`, `wait_all`, `wait_any`, `test`, etc.).
This also aligns with MPI semantics w.r.t. non-blocking operations.

src/KokkosComm/mpi/allgather.hpp
dssgabriel: Yes, this corresponds to my question in the PR description:

> - For non-contiguous "receive" interfaces, can we directly unpack into the passed view instead of the host-staged view? This would remove a call to `deep_copy`, which I think is smart enough to do the right thing, but I am not sure. E.g.:
> 
> ```cpp
> auto host_rv = KokkosComm::Impl::stage_for(rv);
> space.fence("fence host staging before `MPI_Recv`");
>
> // Assume non-contiguous view 
> auto packed = Packer::allocate_packed_for(space, "packed `MPI_Recv`", host_rv);
> space.fence("fence packing before `MPI_Recv`");
> MPI_Recv(data_handle(packed.view), packed.count, packed.datatype, src, tag, comm, MPI_STATUS_IGNORE);
> 
> // NOTE: Can we unpack directly into `rv` instead of `host_rv`
> // and eliminate the subsequent call to `copy_back`?
> Packer::unpack_into(space, host_rv, args.view);
> KokkosComm::Impl::copy_back(space, rv, host_rv);
> 
> space.fence("fence copy back after `MPI_Recv`");

I'll refactor it with a direct `deep_copy` to avoid the (unnecessary) intermediate operation.

PR #211, opened 2026-01-27
feat: add datatype conversion from view utility

dssgabriel:
Add conversion utilities to a CommSpace-associated datatype from a Kokkos View object:
```cpp
using DCS = KokkosComm::DefaultCommunicationSpace;

Kokkos::View<T*> v("v", 10);
auto dtype = KokkosComm::datatype_for<DCS>(v);
// or
auto dtype = KokkosComm::datatype_for(DCS{}, v);
```

Also added the missing unit tests for the `datatype` utility (based on the `red_op_conversion` test).

### DIFF
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -281,3 +281,32 @@ NCCL specialization
 
         Both ``wait_all`` and ``wait_any`` use active polling loops rather than blocking synchronization. While this
         increases CPU utilization, it avoids the overhead of spawning threads or completing requests sequentially.
+
+Utility
+=======
+
+.. cpp:namespace:: KokkosComm
+
+.. cpp:function:: template <CommunicationSpace CommSpace, typename T> auto datatype() -> CS::datatype_type
+
+    Converts a type ``T`` to its communication space ``CS`` equivalent.
+
+    When ``CS`` is:
+
+    * ``MpiSpace``, returns the corresponding ``MPI_Datatype`` type.
+    * ``NcclSpace``, returns the corresponding ``ncclDataType_t`` type.
+
+    :tparam CS: The target communication space backend to use for data type conversion.
+    :tparam T: The type to convert from.
+
+    .. note::
+
+        Non-system data types (i.e. the data types not natively supported by the communication space) are not convertible.
+        This notably includes user-defined types.
+
+.. cpp:function:: template <CommunicationSpace CommSpace, KokkosView V> auto datatype_for(V&) -> CS::datatype_type
+
+    Returns the ``CS`` data type equivalent for the value type of the Kokkos View ``V``.
+
+    :tparam CS: The target communication space backend to use for data type conversion.
+    :tparam V: The Kokkos View to convert the value type from.
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -287,26 +287,38 @@ Utility
 
 .. cpp:namespace:: KokkosComm
 
-.. cpp:function:: template <CommunicationSpace CommSpace, typename T> auto datatype() -> CS::datatype_type
+.. warning::
 
-    Converts a type ``T`` to its communication space ``CS`` equivalent.
+    Non-system data types (i.e. the data types not natively supported by the communication space) are not convertible.
+    This notably includes user-defined types.
 
-    When ``CS`` is:
+.. cpp:function:: template <CommunicationSpace C, typename T>\
+                  auto datatype() -> C::datatype_type
+
+    Converts a type ``T`` to its communication space ``C`` equivalent representation.
+
+    When ``C`` is:
 
     * ``MpiSpace``, returns the corresponding ``MPI_Datatype`` type.
     * ``NcclSpace``, returns the corresponding ``ncclDataType_t`` type.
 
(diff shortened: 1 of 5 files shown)


cedricchevalier19 (dismissed):

cedricchevalier19 (commented):
Are you able to reproduce the CI error on your side?

dssgabriel:
I haven't tested yet with the CUDA + OMPI setup. Will do tomorrow morning on the work laptop.

However, I am considering removing the datatype conversion for "any" generic type entirely since we have no use cases for it... (contrary to converting from a View, hence this PR).

dssgabriel:
Is the runner for the compile-only job broken? I don't understand why the compilation fails on something related to Google Tests since we haven't touched any of that.

cedricchevalier19:
It is possible that we have an issue between the newer Kokkos and GTest. I'll try to have a look on that next Monday.

dssgabriel:
@cedricchevalier19 I have fixed the issue.
The errors were triggered because some GTest assertions were not wrapped in braces (single-line bodies of `if` statements checking `size_t` width).
We can merge!

src/KokkosComm/datatype.hpp
```diff
@@ -152,9 +152,9 @@ template <CommunicationSpace CS, typename T>
   }
 }
 
-template <CommunicationSpace CS, typename T>
-[[nodiscard]] constexpr auto datatype_for(T&&) -> typename CS::datatype_type {
-  return datatype<CS, std::remove_cvref_t<T>>();
+template <CommunicationSpace CS, KokkosView V>
+[[nodiscard]] constexpr auto datatype_for(const V&) -> typename CS::datatype_type {
+  return datatype<CS, std::remove_cvref_t<typename V::value_type>>();
```
cedricchevalier19: I think this function should be documented:
- Add a comment in the source
- Add it in the doc?

src/KokkosComm/datatype.hpp
cedricchevalier19: Drive-by, but is is ok even without MPI?

dssgabriel: > is it ok even without MPI?

What do you mean?
If neither MPI nor NCCL backends have been enabled, the compilation would fail earlier than this (we have an `#error` PP directive that is triggered if no communication backend has been enabled). The `else` statement in this function is unreachable.

cedricchevalier19: So if `NCCL` is enabled and not the `MPI` backend, we do not have problems?

dssgabriel: Yes, this is fine because NCCL implicitly declares the `MpiSpace` as a fallback, see https://github.com/kokkos/kokkos-comm/blob/2719002985bbbf6c6c6a9f78c26d2cf9cdcd24b1/src/KokkosComm/fwd.hpp#L13-L18
So the `MPI` backend is technically always available.

docs/api/core.rst
```diff
@@ -111,3 +111,33 @@ Related types
...
+    * ``MpiSpace``, returns the corresponding ``MPI_Datatype`` type.
+    * ``NcclSpace``, returns the corresponding ``ncclDataType_t`` type.
+
+    :tparam CS: The target communication space backend to use for data type conversion.
+    :tparam T: The type to convert from.
+
+    .. note::
+
+        Non-system data types (i.e. the data types not natively supported by the communication space) are not convertible.
+        This notably includes user-defined types.
```
cedricchevalier19: So what is returned in this case?

src/KokkosComm/datatype.hpp
```diff
@@ -138,6 +138,14 @@ constexpr auto nccl_datatype() -> ncclDataType_t {
...
 }  // namespace Impl
 
+/// Converts a type `T` to its communication space `CS` equivalent.
+///
+/// When `CS` is:
+/// - `MpiSpace`, returns the corresponding `MPI_Datatype` type.
+/// - `NcclSpace`, returns the corresponding `ncclDataType_t` type.
+///
+/// Non-system data types (i.e. the data types not natively supported by `CS`) are not convertible. This notably
+/// includes user-defined types.
```
cedricchevalier19: Same here.

cedricchevalier19 (dismissed):
I do not ask to magically deal with custom datatype but their handling should be defined in the documentation.

For later, I am still thinking that using `std::optional` for this kind of query is a good thing.

docs/api/core.rst
dssgabriel: Compilation fails on a `static_assert` :(

dssgabriel:
> For later, I am still thinking that using std::optional for this kind of query is a good thing.

Do we want to refactor this with `std::optional` before merging? I don't think this is a high-priority merge, although it would be useful for refactoring & harmonizing the MPI bindings in preparation for host staging.

docs/api/core.rst
cedricchevalier19: So please add this in the documentation.

dssgabriel:
@cedricchevalier19 I clarified the docs, so it should be ready to merge now. Do we want to pass SNL CI first, though?

cedricchevalier19 (approved):

PR #216, opened 2026-02-10
refactor: rename and rework `Request`s impls

dssgabriel:
## Summary

This PR comprehensively refactors the request handling implementation across both MPI and NCCL backends, modernizing the API and improving type safety, error handling, and resource management.

## Motivation

The previous request implementation had several limitations:
- Requests relied on `shared_ptr` wrapping due to copyable semantics
- Inconsistent API patterns between MPI and NCCL backends
- Limited error handling
- Dead code and unused abstractions

### Core Request Redesign

- **Name**: Renamed ~~`Req`~~ -> `Request`
- **Move-only semantics**: Requests can no longer be copied, eliminating `shared_ptr` wrapping overhead
- **Unified API**: Consistent member functions across backends (`request()`, `request_ptr()`, `wait()`, `test()`)
- **Simplified callbacks**: Renamed various callback functions -> `add_callback()`

### MPI Backend

- Eliminated internal `Record` struct, removed dead code
- Fixed `wait_all` and `wait_any` implementations
- Improved error handling
- Renamed `req.hpp` -> `request.hpp`

### NCCL Backend

- Changed internal handle: `cudaStream_t` -> `cudaEvent_t` for more precise synchronization
- Eliminated internal `Record` struct, removed dead code
- Reworked `wait_all` with active-wait polling instead of a sequential loop
- Consolidated error handling utilities into `impl/error_handling.hpp`

### Propagation

- Updated core API and all MPI/NCCL collective and point-to-point operations
- Updated documentation and tests

## Breaking Changes

- Requests can no longer be copied, only moved
- API function names changed: `mpi_request()` / `get_inner()` → `request()`
- Callback registration function renamed to `add_callback()`

## Testing

All existing unit tests and performance tests have been updated to work with the new request API.


dssgabriel:
That's weird :thinking:
On macOS, the Open MPI workflow fails on compilation because `MPI_REQUEST_NULL` is not constexpr, but on Linux both workflows (`Linux / Kokkos-master-OpenMPI-OpenMP-Release` and the MPI backend of `Compile-only / Kokkos_master-Open_MPI_508-CUDA_1290-Release`) pass fine...

OS | Open MPI version
--- | ---
macOS | 5.0.9
Ubuntu 24.04 | 4.1.6 

Edit: could be because of Apple Clang's compiler being the default and not handling that

cedricchevalier19:
> On macOS, the Open MPI workflow fails on compilation because `MPI_REQUEST_NULL` is not constexpr,

We already had this issue. It can happen even on linux, and we should not rely on constrexpr-ness of something from C.

src/KokkosComm/mpi/request.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+  /// @brief Queries the request for the completion of the associated operation.
+  /// If the operation has completed, all callbacks are executed and the underlying `MPI_Request` object is set to
+  /// `MPI_REQUEST_NULL` upon return, similarly to having called `wait`.
+  /// @return True if the request has completed, is null, or inactive, false otherwise.
+  [[nodiscard]] auto test() -> bool {
+    if (not is_active()) {
+      return true;
+    }
+
```
cedricchevalier19: I think MPI_Test has the same behavior.

src/KokkosComm/mpi/request.hpp
```diff
@@ -0,0 +1,189 @@
...
+  /// @return True if the request is active, false otherwise.
+  [[nodiscard]] auto is_active() const noexcept -> bool { return request_ != MPI_REQUEST_NULL; }
+
+  /// @brief Waits on the request until completion of the associated operation.
+  /// The underlying `MPI_Request` object is set to `MPI_REQUEST_NULL` upon return.
+  auto wait() -> void {
+    if (not is_active()) {
+      return;
+    }
+
```
cedricchevalier19: Not sure we need this. I think it is correct to call `MPI_Wait` on `MPI_REQUEST_NULL`.

cedricchevalier19 (commented):
I like the refactoring.

dssgabriel:
> https://github.com/kokkos/kokkos-comm/pull/216/changes#diff-6884d949d0a3d846c378c4f001c2764f8fcde8b874c36bbd810fdeb5e690cc96R75-R77
> Not sure we need this. I think it is correct to call MPI_Wait on MPI_REQUEST_NULL.

> https://github.com/kokkos/kokkos-comm/pull/216/changes#diff-6884d949d0a3d846c378c4f001c2764f8fcde8b874c36bbd810fdeb5e690cc96R92-R94
> I think MPI_Test has the same behavior.

Should I just remove the `is_active` helper then? The point was to have a "quicker" function to check the Request's status that relies on invariants and avoids calling the full-blown `MPI_Test`. However, if we don't use it in our own implementations, we should constrain the interface and eliminate it from the API.

dssgabriel:
Success! This should be ready for review @cedricchevalier19

src/KokkosComm/mpi/irecv.hpp
```diff
@@ -20,23 +20,23 @@ namespace Impl {
...
+                src, POINTTOPOINT_TAG, h.mpi_comm(), req.request_ptr());
       req.extend_view_lifetime(rv);
     } else {
       auto args = Packer::allocate_packed_for(space, "TODO", rv);
       space.fence("fence before irecv");
-      MPI_Irecv(args.view.data(), args.count, args.datatype, src, POINTTOPOINT_TAG, h.mpi_comm(), &req.mpi_request());
+      MPI_Irecv(args.view.data(), args.count, args.datatype, src, POINTTOPOINT_TAG, h.mpi_comm(), req.request_ptr());
       // implicitly extends args.view and rv lifetime due to lambda capture
-      req.call_after_mpi_wait([=]() { Packer::unpack_into(space, rv, args.view); });
+      req.add_callback([=]() { Packer::unpack_into(space, rv, args.view); });
```
cedricchevalier19: Perhaps we can be explicit and list captured elements?

src/KokkosComm/nccl/request.hpp
```diff
@@ -0,0 +1,197 @@
...
+
+  /// @brief Capture the state of a `cudaStream_t` for request encapsulation.
+  /// @param stream The stream to capture for request encapsulation.
+  auto capture_stream_state(cudaStream_t stream) noexcept -> void {
+    if (request_ != nullptr) {
+      KC_CUDA_CHECK(cudaEventDestroy(request_));
+    }
+    KC_CUDA_CHECK(cudaEventCreate(&request_, cudaEventDisableTiming));
+    KC_CUDA_CHECK(cudaEventRecord(request_, stream));
+  }
```
cedricchevalier19, line 40: Do we need this as public?

cedricchevalier19 (dismissed):
You should add somewhere (perhaps a developer doc?, or in the comment of Request implementations) why a move semantic is interesting to ensure that callbacks are called only once.

src/KokkosComm/nccl/request.hpp
dssgabriel: Unfortunately, yes. Otherwise, we can't capture the stream state at the right moment in the NCCL bindings.

The only way to do without it would be to create the request and return it from within all the nested code paths, which would produce some code duplication.

E.g., for `nccl::recv` we are doing the following, with the common logic outside of the `if`/`else` statements:
```cpp
Request<NcclSpace> req;

if (is_contiguous(rv)) {
  ncclRecv(data_handle(rv), span(rv), datatype<NcclSpace, T>(), peer, comm, space.cuda_stream());
} else {
  using Packer = typename Impl::PackTraits<RecvView>::packer_type;
  auto pckd_rv = Packer::allocate_packed_for(space, "pckd_rv", rv);
  ncclRecv(data_handle(pckd_rv.view_), pckd_rv.count_, pckd_rv.datatype_, peer, comm, space.cuda_stream());
  req.capture_stream_state(space.cuda_stream());
  req.add_callback([=]() {
    Packer::unpack_into(space, rv, pckd_rv.view_);
    space.fence("fence `pckd_rv` unpacking after NCCL call");
  });
}

req.extend_view_lifetime(rv);
return req;
```
And we would need to do the following, moving that common logic inside both branches and thus duplicating some LOCs:
```cpp
if (is_contiguous(rv)) {
  ncclRecv(data_handle(rv), span(rv), datatype<NcclSpace, T>(), peer, comm, space.cuda_stream());
  Request<NcclSpace> req(space.cuda_stream());
  req.extend_view_lifetime(rv);
  return req;
} else {
  using Packer = typename Impl::PackTraits<RecvView>::packer_type;
  auto pckd_rv = Packer::allocate_packed_for(space, "pckd_rv", rv);
  ncclRecv(data_handle(pckd_rv.view_), pckd_rv.count_, pckd_rv.datatype_, peer, comm, space.cuda_stream());
  Request<NcclSpace> req(space.cuda_stream());
  req.extend_view_lifetime(rv);
  req.add_callback([=]() {
    Packer::unpack_into(space, rv, pckd_rv.view_);
    space.fence("fence `pckd_rv` unpacking after NCCL call");
  });
  return req;
}
```

dssgabriel:
Before merging, I will go ahead and fully update the documentation regarding the new Requests

src/KokkosComm/nccl/request.hpp
cedricchevalier19: Thanks for the explanation.

cedricchevalier19 (approved):

PR #217, opened 2026-02-19
chore: cleanup traits + improve docs for concepts and traits

### LINKED ISSUE #200: Review public API
- [ ] Ensure it is "minimal", i.e., we do not expose to much implementation details
- [ ] Ensure it is consistent in between the modules
- [ ] Ensure it is consistent with Kokkos

### LINKED ISSUE #201: Update Documentation
Issue to centralize missing documentation.

dssgabriel:
Cleanup trait functions:
- Added `size` trait
- Added overloads taking a (unused) View param for `rank` and `is_reference_counted`, so one can write, e.g.:
  `rank(view)` instead of `rank<V>()`
- `constexpr`-ify everything that could be
- Decorate the interfaces with all the necessary attributes/qualifiers/operators (`[[nodiscard]]`, `[[maybe_unused]]`, `noexcept`, etc. stuff)
- Document the code for better LSP preview in text editors/IDEs
- Update the matching docs
  - Split concepts into their own page
  - Better document concepts and bring them up to date with regards to requirements/constraints
  - Document reduction operators tag-like structures
  - Update traits docs with cleaned up interfaces

### DIFF
--- a/docs/api/concepts.rst
+++ b/docs/api/concepts.rst
@@ -1,0 +1,130 @@
+********
+Concepts
+********
+
+.. cpp:namespace:: KokkosComm
+
+Kokkos Comm-specific concepts
+=============================
+
+.. cpp:concept:: template <typename T> CommunicationSpace
+
+    Specifies that a type ``T`` is a Kokkos Comm communication backend.
+
+    **Constraints for implementing the ``CommunicationSpace`` concept**
+
+    .. cpp:type:: T::communication_space
+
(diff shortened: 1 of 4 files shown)


cedricchevalier19 (approved):

dssgabriel:
SNL CI can't pass anymore since the CMake on there is 3.23, and we require 3.25...

Since these are mostly small changes + docs, and every other job is passing, I think we can merge as is.

PR #218, opened 2026-02-20
ci: install minimum supported version of CMake on SNL CI jobs

dssgabriel:
Install CMake 3.25.3 on SNL CI jobs so they can actually compile.


cedricchevalier19:
@cwpearson can you explicitly relaunch the SNL CI?

cwpearson:
Do we need this? I updated the runner image to include cmake 3.25.3 a while ago.

cedricchevalier19:
In this case, we can close this.

PR #220, opened 2026-03-05
refactor!: rename and rewrite `Handle` impls

dssgabriel:
### Description

Similar to the recent work on requests (#216), this PR attempts to rename and rewrite the implementations of `Handle`s, now known as `Communicator`s, in view of stabilizing the Kokkos Comm API for our first official release.

- **Related issue(s):** #200, #201
- **Depends on:** none

<details>

**Motivation**
These changes attempt to stabilize Kokkos Comm's public API for "handles". In particular, it aims at:
- Settling on a "correct" name for these objects that wrap a library-specific communicator handle and a Kokkos execution space
- Removing any backend-dependent interfaces from their public API
- Expanding the capabilities of these objects
- Defining a set of mandatory supported operations
- Unifying implementation strategies in the backend specializations (as much as possible)
- Bringing the documentation up-to-date

**Implementation notes**

__Unified interface across backends:__
The central design constraint was that `Communicator<Ex, MpiSpace>` and `Communicator<Ex, NcclSpace>` must expose identical public interfaces, so that code written against the abstraction does not need to branch on the backend.

__Factory functions instead of constructors:__
Construction is intentionally not possible via public constructors because there are multiple ways to create communicators. Static "factory" member functions (`from_raw`, `duplicate_from_raw`, `split_from_raw`) were chosen for three reasons:
1. They have meaningful names that make construction intent explicit at the call site;
2. They can return `std::optional` to force callers to handle failure without exceptions; and
3. They make it impossible to accidentally construct a `Communicator` in an invalid state.

A builder pattern was considered but rejected — with only three construction modes and minimal configuration, a builder would add object lifetime complexity for little real benefit, and in the long run, could promote backend-specific setter methods that would gradually erode the generic interface.

__Non-owning `from_raw`:__
`from_raw` never takes ownership of the raw handle it is given. If a user created a raw communicator, they opted out of the abstraction layer and retain responsibility for its lifetime. Allowing optional ownership transfer would reintroduce a boolean parameter at the call site and recreate the exact class of double-free and leak bugs the design is meant to prevent. The ownership table is simple and memorable: only `duplicate` and `split` produce owned communicators, because only they allocate a new backend handle internally.

__Explicit copies, move-only by default:__
The copy constructor and copy assignment operator are deleted. Duplicating a communicator is an explicit MPI/NCCL operation with observable side effects and a potential for failure; silently copying a `Communicator` as though it were a value type would hide that cost entirely. Users who need a copy call `duplicate()`, which makes the intent and the cost visible. Move semantics are implemented manually rather than defaulted: the move constructor uses `std::exchange` to null out the source handle and set `owned_` to `false`, ensuring the moved-from object's destructor is always a no-op and preventing double-free.

</details>

### Changes

- **Affected areas:** all
- **Breaking change(s)?** yes

- Renamed ~~`Handle`~~ -> `Communicator` across the entire code base, including file names, type aliases, and documentation
  - After discussing with potential users during the 2026 HPSF Community Summit, this is the name that seemed to be the most well understood by the community
- Rewrote both MPI and NCCL `Communicator` specializations with a unified interface
  - Added `duplicate_from_raw` and `split_from_raw` factory operations for both backends, each accepting either a raw backend handle. Also support duplicating/splitting from an existing `KokkosComm::Communicator` using `duplicate` and `split` member functions.
  - Added `from_raw` factory function, which borrows a raw backend handle without taking ownership
  - Unified accessor interface across both specializations: `exec()`, `comm()`, `size()`, `rank()`
  - Deleted copy constructor and copy assignment operator; use `duplicate()` for explicit copies
  - Implemented correct move constructor and move assignment operator
- Refactored all existing tests to use the new `Communicator` API
- Added a dedicated unit test suite for `Communicator`, covering factory functions, member functions, accessors, and move semantics for both MPI and NCCL backends
- Rewritten documentation fully covering the new `Communicator` object
- Updated `clang-format` config to avoid bin-packing arguments, improving readability of functions with long signatures (especially factory functions with attributes, default values, etc.)

**Drive-by changes**
- Enabled previously forgotten broadcast and all-gather tests (files existed but weren't declared in CMake)

### Checklist

- [x] Tests are up-to-date


dssgabriel:
I cannot test on a NCCL system right now, but everything passes for both Open MPI (5.0.9) and MPICH (5.0.0) for the MPI backend (non-CUDA enabled) on my machine.

dssgabriel:
@cwpearson Can you enable the SNL CI so I can fix any outstanding issues before making this ready for review?

docs/api/core_recv.cpp
```diff
@@ -1,24 +1,24 @@
-#include "KokkosComm/KokkosComm.hpp"
+#include <Kokkos_Core.hpp>
+#include <KokkosComm/KokkosComm.hpp>
 
-// Define the execution space and transport
-using ExecSpace = Kokkos::DefaultExecutionSpace;
-using CommSpace = DefaultCommunicationSpace;
+// Define the communication and execution spaces
+using Co = KokkosComm::DefaultCommunicationSpace;
+using Ex = Kokkos::DefaultExecutionSpace;
```
cedricchevalier19: Not blocking but aliasing is not easy to read later. Are something a little bit longer like `Comm` or `Exec`.

docs/api/core_recv.cpp
```diff
@@ -1,24 +1,24 @@
...
 
 // Source rank
-int src = 1;
+int src_rank = 1;
 
-// Create a handle
-KokkosComm::Handle<> handle; // Same as Handle<Execspace, CommSpace>
+// Create a communicator
+auto comm = KokkosComm::Communicator<Co, Ex>::duplicate(raw_comm_handle, exec_space);
```
cedricchevalier19: For later, we should deduce the correct template parameters from the arguments. Or do we make a function that duplicates a Communicator.

src/KokkosComm/mpi/communicator.hpp
```diff
@@ -0,0 +1,144 @@
...
+  using size_type           = communication_space::size_type;
+  using rank_type           = communication_space::rank_type;
+
+  /// @brief Constructs a `Communicator` from a raw `MPI_Comm` handle and a Kokkos execution space instance.
+  /// Defaults `exec` to `Kokkos::DefaultExecutionSpace`.
+  /// Passing `MPI_COMM_NULL` returns `nullopt`.
+  ///
+  /// The passed `comm` must be a valid handle and must not be an inter-communicator parent handle.
+  /// The returned communicator does not own the underlying handle, and the user may be responsible for destroying it.
+  /// Therefore, `MPI_COMM_WORLD` and `MPI_COMM_SELF` are valid communicator handles to pass.
```
cedricchevalier19: I think it is ok to duplicate the given communicator and therefore to handle the destruction of our object.

cedricchevalier19 (commented):
I am wondering if we can get rid of the non owning aspect for the Communicator.
The code and the semantics may be simpler.

The other thing is the new `Communicator` should be used by all public functions of Kokkos Comm that use a communicator. So even the advanced functions in `::mpi::`, should not take a `MPI_Comm`.

docs/api/core_recv.cpp
dssgabriel: I am changing this :+1:

docs/api/core_recv.cpp
dssgabriel: Unfortunately, we cannot deduce without taking an `MpiSpace` tag as argument, since we cannot define CTAD-like rules for static member functions.
An alternative could be the free function you mention, but I don't know about that. The name would need to reflect that it operates on a communicator, which isn't as clean as having a static member function.

One easy improvement we can make is to default the template parameters of the Communicator:
```cpp
template <
  CommunicationSpace Comm = DefaultCommunicationSpace,
  KokkosExecutionSpace Exec = Kokkos::DefaultExecSpace>
struct Communicator;
```
which allows for simpler "default" code:
```cpp
auto comm = KokkosComm::Communicator<>::duplicate(raw_comm_handle, exec);
```

dssgabriel:
> I am wondering if we can get rid of the non-owning aspect for the Communicator.
The code and the semantics may be simpler.

TLDR: I think we should keep it for now. 

I tend to agree, but this makes it harder for existing MPI-based codes to enter the Kokkos Comm "ecosystem". Namely, for libraries or apps that already manage their own MPI communicator (create/dup/split + free), having the `from_raw` static member function makes it very easy to tap into Kokkos Comm with minimal changes. As already discussed at length, having strong interoperability with MPI-native objects is one of our goals, and I think `from_raw` helps with that.

Some design suggestions, though:
- Make it even simpler and not return a `std::optional`; what you pass is what you get. This also more clearly separates this function from `duplicate` and `split`, and "accentuates" the fact that we are not responsible for managing this communicator.
- Mark it `[[deprecated("Prefer creating Communicators with duplicate() or split() instead")]]`, so we can plan its removal from the public API at some point (e.g., once we have sorted out initialization and can do everything ourselves)?

---
> The other thing is that the new Communicator should be used by all public functions of Kokkos Comm that use a communicator. So even the advanced functions in `::mpi` should not take an `MPI_Comm`.

I agree. I have other PRs coming that clean up the p2p and colls interfaces (as part of #200) which are in line with this.

dssgabriel:
Tested on my machine with MPI + CUDA and everything works fine.
I'll try to run on a multi-GPU system so I can test NCCL while we wait for SNL's CI.

dssgabriel:
I added a few more unit tests for Communicators and got them to run with the NCCL backend on a GPU system.

This is ready for review and merge @cedricchevalier19.
@cwpearson, can you enable the SNL CI, please?

docs/api/core_send.cpp
```diff
@@ -1,29 +1,28 @@
...
+auto req1 = KokkosComm::send(comm, data, dst_rank);
 
-// Initiate a non-blocking send with a default handle
-auto req2 = send(data, dest);
+// Simulate a blocking send by waiting immediately
+KokkosComm::send(comm, data, dst_rank).wait();
 
-// Wait for the requests to complete (assuming a wait function exists)
+// Wait for a request to complete
 KokkosComm::wait(req1);
```
cedricchevalier19, line 28: For later, @hmt23 can you confirm this is also correct from a MPI point-of-view?

cedricchevalier19 (dismissed):
One possible improvement: rename `comm` to `mpi_comm` in the internal MPI API.
Not blocking.

dssgabriel:
> One possible improvement: rename `comm` to `mpi_comm` in the internal MPI API. Not blocking.

I think we can add an overload for it, but not outright rename it, or we will lose the "generic" interface between Communicator specializations (i.e., users will need to know which one they are using if they want to retrieve the raw comm handle)

cedricchevalier19:
> > One possible improvement: rename `comm` to `mpi_comm` in the internal MPI API. Not blocking.
> 
> I think we can add an overload for it, but not outright rename it, or we will lose the "generic" interface between Communicator specializations (i.e., users will need to know which one they are using if they want to retrieve the raw comm handle)

I was thinking about renaming the parameters, no change in the API.

cedricchevalier19 (approved):

PR #221, opened 2026-03-18
test: refactor & add helpers for handling Views of arbitrary rank

dssgabriel:
### Description

Refactored implementation of view builder helpers and added new ones, all now MD-capable:
- Refactored `ViewBuilder` struct into a free function (`build_view`) that allows building views of arbitrary type and dimension, contiguous or not.
- Added view initialization (`init_view`) and error counting (`count_error`) helpers.

These changes make it possible to write only the high-level test function and pass different kinds of Views to it. Such tests better mirror user code that Kokkos Comm is designed to handle correctly.
It is now trivial to test views of higher dimensions (>2), contiguous or not (provided we support them)*.

<details>
<summary>*Note (rant)</summary>

We currently don't support non-contiguous Views of rank higher than 2 with the DeepCopy packing strategy due to limitations stemming from a similar implementation approach to the one refactored in this PR.

</details>

In (a) follow-up PR(s), all tests should be refactored to use this new mechanism. Additionally, this will let us easily enhance existing tests to actually assert the expected behavior of all the functionalities Kokkos Comm is designed to provide.

- **Related issue(s):** none
- **Depends on:** #220 

<details>
<summary><b>Additional context</b></summary>

**Motivation**
Initializing views and counting errors were the two problematic pieces of code that forced us to have dedicated test functions for each view dimension we wanted to test. The new helpers abstract away the required MD logic, letting us write simpler tests that are much more representative of what users would write.

</details>

### Changes

- **Affected areas:** unit tests
- **Breaking change(s)?** no

Details:
- Refactored `ViewBuilder` struct into a free function for building views of arbitrary type and dimension (contiguous or not)
- Added view initialization and error counting helpers (MD-capable)
- Wrapped the helpers in a `test_utils::` namespace
- Removed redundant `view_builder.hpp` and consolidated logic into `view_utils.hpp`
- Updated dependent tests in `unit_tests/test_sendrecv.cpp`

### Checklist

- [x] Tests are up-to-date


dssgabriel:
Replaced by #224

PR #222, opened 2026-03-25
feat(core): add contiguous packing support for Views of arbitrary rank

dssgabriel:
### Description

Refactors the contiguous packing infrastructure to add support for Views of any dimensions using variadic templates, replacing the previous implementation, which limited us to 2D Views at most.

- **Related issue(s):** none (but mentioned in PR #221)
- **Depends on:** #224

<details>

</details>

### Changes

- **Affected areas:** core
- **Breaking change(s)?** no

- Add support for Views of arbitrary rank in `allocate_contiguous_for` via `std::index_sequence` and parameter pack expansion refactoring, replacing `if constexpr` branches for rank 1 and 2
- Removed unused `resize_contiguous_for` interface (not used anywhere in the Kokkos Comm code base)
- Make View label an optional parameter in `allocate_contiguous_for` (use a default value if omitted). This change does not affect any user-facing interfaces and is just a convenience in preparation for a larger refactoring of non-contiguous handling strategies.
- Update MPI and NCCL backend packers to use the new `allocate_contiguous_for` signature
- Add unit tests asserting the behavior of `allocate_contiguous_for` on Views of rank 0 to rank 7 (rank-8 non-contiguous View)

### Checklist

- [x] Tests are up-to-date


src/KokkosComm/impl/contiguous.hpp
```diff
@@ -21,31 +23,36 @@ struct contiguous_view {
...
+/// @brief Allocate a contiguous View suitable for packing a non-contiguous View.
+/// @tparam Exec A Kokkos Execution Space type.
+/// @tparam View A Kokkos View type.
+/// @param exec The execution space instance in which to perform the view allocation.
+/// @param v The View to make a suitable contiguous allocation for.
+/// @param label The label to give to the allocated contiguous View. Defaults to "contiguous_view".
+template <KokkosExecutionSpace Exec, KokkosView View>
+auto allocate_contiguous_for(const Exec& exec, const View& v, const std::string& label = "contiguous_view") {
+  using ContigView = contiguous_view_t<View>;
+  return [&]<size_t... Is>(std::index_sequence<Is...>) {
```
cedricchevalier19: Can you be explicit about the capture?

src/KokkosComm/impl/contiguous.hpp
```diff
@@ -21,31 +23,36 @@ struct contiguous_view {
...
+/// @tparam DstV The Kokkos View type of the destination View to resize.
+/// @tparam DstV The Kokkos View type of the source View to resize for.
+/// @param exec The execution space instance in which to perform the view reallocation.
+/// @param dst The View to resize.
+/// @param src The View to make a suitable contiguous resize for.
+template <KokkosExecutionSpace Exec, KokkosView DstV, KokkosView SrcV>
+auto resize_contiguous_for(const Exec& exec, const DstV& dst, const SrcV& src) {
+  static_assert(rank<DstV>() == rank<SrcV>(),
+                "KokkosComm::Impl::resize_contiguous_for: source and destination Views must have the same rank");
+  [&]<size_t... Is>(std::index_sequence<Is...>) {
```
cedricchevalier19: Can you be explicit about the capture?

src/KokkosComm/impl/contiguous.hpp
cedricchevalier19: I am not sure that this compact code is very readable. The other option will be to give a name to the lambda and to call it.

src/KokkosComm/mpi/impl/packer.hpp
```diff
@@ -35,7 +35,7 @@ struct DeepCopy {
   /// Returns allocated, uninitialized, contiguous view for packing `src`.
   template <KokkosExecutionSpace ES>
   static auto allocate_packed_for(const ES &space, const std::string &label, const V &src) -> Args {
-    auto packed = KokkosComm::Impl::allocate_contiguous_for(space, label, src);
+    auto packed = KokkosComm::Impl::allocate_contiguous_for(space, src, label);
```
cedricchevalier19: In Kokkos, label is often the first parameter.

cedricchevalier19 (commented):
No blockers but some questions.

src/KokkosComm/mpi/impl/packer.hpp
dssgabriel: I am reverting this.

src/KokkosComm/impl/contiguous.hpp
dssgabriel: This IILE pattern is a common idiom for iterating over a pack (here, the indexes produced by `std::index_sequence`). I would rather add an explanatory comment than make it a named lambda.

cedricchevalier19: No, it is fine.

src/KokkosComm/mpi/impl/packer.hpp
cedricchevalier19: It means it is before `space` also.

dssgabriel: Good to know, thanks! The previous interface wasn't even very Kokkos-like then 😅

dssgabriel: I simply reverted it to what it was; I would have had to propagate the change to the backend-specific `Packer`s and to the primitives with non-contiguous support as well, which would have made it an unnecessarily large change.

cedricchevalier19 (commented):
Can you add a small unittest for that?

dssgabriel:
Yes, good point. I'll be able to test if this change handles rank-0 Views.

cwpearson (dismissed):

dssgabriel:
@cwpearson I removed the `resize_contiguous_for` API, which was not used anywhere in the Kokkos Comm codebase.
Is this blocking for you in any way? I suppose not, since this was behind `Impl::`, but I figured I should call it out.

Otherwise, this will be ready to merge (needs #224 first, though), and should likely supersede #223.

cwpearson (approved):

cwpearson:
I don't see a 0-rank test in here, but regardless this is a better foundation. I'll give it a try in miniFE and see what happens.

dssgabriel:
It is a typed-test on the ranks of the Views; we create non-contiguous Views from rank-0 to rank-7: 
```
10: [==========] Running 8 tests from 8 test suites.
10: [----------] Global test environment set-up.
10: [----------] 1 test from Contiguous/0, where TypeParam = std::integral_constant<unsigned long,0ul>
10: [ RUN      ] Contiguous/0.Allocate
10: [       OK ] Contiguous/0.Allocate (0 ms)
10: [----------] 1 test from Contiguous/0 (0 ms total)
```

PR #223, opened 2026-04-01
Add rank-0 View support to DeepCopy packer

cwpearson:
Many codes (e.g. MiniFE) will make MPI calls against single values, best represented as a rank-0 Kokkos view.

Whether we use the packer is guarded by a runtime check `is_contiguous`. Since it's runtime, the packer will still be instantiated on e.g. rank-0 views, even if the code path is never called.

This PR removes the failure in that case, and also implements a no-op rank 0 path (even though we don't use it).

### DIFF
--- a/docs/api/core_recv.cpp
+++ b/docs/api/core_recv.cpp
@@ -8,7 +8,7 @@ using CommSpace = DefaultCommunicationSpace;
 int src = 1;
 
 // Create a handle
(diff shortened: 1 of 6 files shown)


dssgabriel:
Can you check whether this can be handled by #222?
That PR was basically to enable us to support views of any rank, but I hadn't thought of the rank-0 View case (maybe it's trivially handled?)

cwpearson:
Closed in favor of #222

PR #227, opened 2026-04-14
fix(mpi): correct mpi_request vector size in wait_any

### LINKED ISSUE #123: Missed partial specialization of Barrier for MPI

This should be something like 

```c++

struct Barrier {
    MPI_Barrier(h.mpi_comm());
  }
};

```

dssgabriel:
### Description

The `wait_any` function was incorrectly constructing a `std::vector` with `requests.size()` elements via the size constructor, then pushing back the same number of elements, resulting in a vector of `2*N` elements with the first `N` being default-constructed objects.

Fix by using `reserve()` + `emplace_back()` to ensure only `N` valid `MPI_Request` objects are stored.

---

Pointed out to me by @sbstndb, thanks!

This should be merged before the release since it is a small but important fix.

- **Related issue(s):** none 
- **Depends on:** none 

<details>

</details>

### Changes

- **Affected areas:** mpi 
- **Breaking change(s)?** no 

Changes:
- fix `mpi_request` vector init in `Request<MpiSpace>::wait_any`

### Checklist

- [x] Tests are up-to-date

### DIFF
--- a/src/KokkosComm/mpi/request.hpp
+++ b/src/KokkosComm/mpi/request.hpp
@@ -152,9 +152,10 @@ inline auto wait_any(std::span<Request<MpiSpace>> requests) -> std::optional<typ
     return std::nullopt;
   }
 
-  std::vector<MPI_Request> mpi_requests(requests.size());
+  std::vector<MPI_Request> mpi_requests;
+  mpi_requests.reserve(requests.size());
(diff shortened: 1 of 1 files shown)


cedricchevalier19 (approved):
LGTM

Good catch

PR #245, opened 2026-06-16
feat: mutable view concept

### LINKED ISSUE #123: Missed partial specialization of Barrier for MPI

dssgabriel:
### Description

This PR introduces the `MutKokkosView` concept.
This allows distinguishing between Views that are meant to be updated ("receive Views") vs. Views that are simply read ("send Views").

This improves the robustness of Kokkos Comm’s type system by encoding communication buffer directionality in the template types: send views may be mutable or const, while receive/in-place views must be mutable at compile time. This prevents invalid const-value-typed Views from reaching backend calls (which we were not properly catching prior to this change) and makes the API contract visible in template signatures.

I chose to go with `MutKokkosView` (which is kinda Rust-like) instead of the more natural `ConstKokkosView`, because enforcing "constness" on send Views' value type may force users to const-cast their View's value type upon each Kokkos Comm call, introducing unnecessary verbosity. Besides, we would still need to ensure that receive Views are writable.

This change also prepares for the staging + packing API refactor I am tidying up. 

- **Related issue(s):** none 
- **Depends on:** none 

### Changes

- **Affected areas:** core, backends, docs 
- **Breaking change(s)?** no 

Changes:
- Added `MutKokkosView` concept + its docs
- Replaced `KokkosView` with `MutKokkosView` for "receive Views" in all interfaces.

### Checklist

- [x] Tests are up-to-date

### DIFF
--- a/docs/api/concepts.rst
+++ b/docs/api/concepts.rst
@@ -130,6 +130,19 @@ Kokkos interoperability concepts
 
     Specifies that a type ``T`` is a ``Kokkos::View`` object.
 
+    This concept accepts both mutable views, such as ``Kokkos::View<T*>``, and views with a constant value type, such as ``Kokkos::View<const T*>``.
+
+    Kokkos Comm uses this concept for view arguments that are not written by the communication operation, such as send buffers.
+    The view may still be mutable for other uses, such as Kokkos kernels.
+
+
+.. cpp:concept:: template <typename T> MutKokkosView
+
+    Specifies that a type ``T`` is a mutable ``Kokkos::View`` object, such as
+    ``Kokkos::View<T*>``.
+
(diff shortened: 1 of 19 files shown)


cedricchevalier19 (approved):
Thanks @dssgabriel, I think it is a great addition. We were previously thinking about `ViewIn` `ViewOut` but I think  `MutKokkosView` is better.

I agree that it is quite a rusty approach but my opinion is that we should give it a try and share the results with other Kokkos projects.

However I am curious about the feelings of @dalg24, @crtrott and @lucbv about this.

cwpearson (approved):

PR #248, opened 2026-06-20
Add core collectives documentation

cwpearson:


dssgabriel (approved):
For another PR, we should consider splitting the core API docs into multiple files to make them less dense for users.

PR #249, opened 2026-06-23
document reduction operator conversion

cwpearson:


dssgabriel (changes requested):
This should be moved to [the end of the `docs/api/core.rst`](https://github.com/kokkos/kokkos-comm/blob/develop/docs/api/core.rst#L561-L600), alongside `datatype` and `datatype_for` utilities, instead of documented in `concepts.rst`.

LGTM otherwise 👍

dssgabriel (approved):

PR #261, opened 2026-07-04
fix(nccl): cudart as build dependency

### LINKED ISSUE #260: Explicitly require CUDA runtime when NCCL backend is enabled
Kokkos Comm should explicitly require the CUDA runtime when the NCCL backend is enabled.

While preparing the Spack package recipe, we found that Kokkos Comm cannot be built with NCCL support unless CUDA is also declared as a build dependency. The NCCL backend relies directly on the CUDA runtime headers (`cuda_runtime.h`) and objects/functions (`cudaStream_t`, `cudaEvent_t`, etc.), so the dependency needs to be explicitly represented in Kokkos Comm’s build rather than inferred indirectly through Kokkos or NCCL.

Reference: https://github.com/spack/spack-packages/pull/5078/changes/10e841086a89e18b575c17037108b63547411400

**Expected behavior**

When `KokkosComm_ENABLE_NCCL=ON`, Kokkos Comm should:

- find the CUDA Toolkit (runtime) dependency explicitly;
- link/export the CUDA runtime target needed by NCCL backend headers (as well as any tests needing CUDART headers);
- ensure that installed CMake package consumers receive the same dependencies.

dssgabriel:
### Description

This PR makes the CUDA runtime an explicit dependency when building Kokkos Comm with the NCCL backend enabled. The NCCL backend includes CUDA runtime headers and uses CUDA runtime types/functions directly, so the build and installed CMake package needs to expose that dependency instead of relying on it indirectly through Kokkos or NCCL.

- **Related issue(s):** closes #260
- **Depends on:** none

### Changes

- **Affected areas:** build, nccl, tests, docs
- **Breaking change(s)?** no

List:
- Require `CUDAToolkit` when `KokkosComm_ENABLE_NCCL=ON`.
- Link the NCCL backend interface against `CUDA::cudart`.
- Propagate the CUDA runtime dependency through the installed `KokkosCommConfig.cmake`.
- Ensure direct NCCL test targets link against `CUDA::cudart`.
- Document the CUDA Toolkit runtime requirement for the NCCL backend.
- Replace direct `cuda.h` inclusion with `cuda_runtime.h`.
- Build the MPI CUDA smoke test as a `.cu` source with CUDA language support.
- Use `CUDA_RUNTIME_LIBRARY Shared` for the MPICH CUDA smoke-test workaround.

### Checklist

- [x] Tests are up-to-date


cedricchevalier19 (approved):
Looks ok

PR #263, opened 2026-07-16
docs: some guidelines for contributing to docs

cwpearson:
Partial fix for https://github.com/kokkos/kokkos-comm/issues/201


dssgabriel (approved):
Maybe we should consider solutions like nix flake to provide a developer environnement that isn't just pre-commit. clang-tidy and other static analysis could be provided within that env too.

PR #269, opened 2026-09-02
fix(mpi/tests): use `MpiSpace` instead of the default

### LINKED ISSUE #123: Missed partial specialization of Barrier for MPI

dssgabriel:
### Description

Use `MpiSpace` explictly instead of `DefaultCommunicationSpace` in MPI-specific tests.

This would lead to compilation errors in builds that enabled both MPI and NCCL backends, because the default comm space would be `NcclSpace` in tests that expected MPI-backed communicators.

- **Related issue(s):** none 
- **Depends on:** none 

### Changes

- **Affected areas:** tests 
- **Breaking change(s)?** no 

### Checklist

- [x] Tests are up-to-date

### DIFF
--- a/unit_tests/mpi/test_isendrecv.cpp
+++ b/unit_tests/mpi/test_isendrecv.cpp
@@ -10,7 +10,7 @@
 namespace {
 
 using Ex = Kokkos::DefaultExecutionSpace;
-using Co = KokkosComm::DefaultCommunicationSpace;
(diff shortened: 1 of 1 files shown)


cedricchevalier19 (approved):
For another PR, we should have tests to check the value of `DefaultCommunicationSpace`.

